PR #9, opened 2025-03-06
chore: Adds info about Lepus VM deprecation in favor of PrimJS VM

DioLps:
So from previous discussion, @pandazyp explained the relationship between Lepus and PrimJS so I'm adding this to the docs.

### DIFF
--- a/docs/en/guide/glossary.mdx
+++ b/docs/en/guide/glossary.mdx
@@ -60,6 +60,16 @@ The thread corresponds to the physical thread regarded as the main thread of the
 
 ### Lepus
 
+<!---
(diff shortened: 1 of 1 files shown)


Huxpro (changes requested):
Oops. It seems like we forgot to remove Glossary.mdx...it is not displayed on anywhere of the website 😅

Our terminologies are defined in the [spec](https://lynxjs.org/guide/spec.html#javascript-engine). If the community feels like we need to explain this word...let's add it back in the speec

DioLps:
Im closing this pr, but we can maybe proceed once Huxpro clarify the questions above.

PR #12, opened 2025-03-06
docs(image): correct mode types description

ArrayZoneYour:
Change-Id: I964faf5f61dc6c867a975eb6d82f25f6d75d1240


cjnhust:
Although the center mode was implemented in the early Lynx code, it had inconsistencies in multi-end rendering effects. We plan to refactor the center mode in future updates. Therefore, it is not currently exposed externally, and we do not recommend its usage.

Huxpro:
Hi @ArrayZoneYour it's great to see you here! 

I'm going to close this, but we'd love to hear more from you :)

PR #13, opened 2025-03-06
fix: Broken Lifecycle Image URL in Markdown

rayfarer:
The image url was pasted twice, resulting in a broken image on the page:

https://lynxjs.org/react/lifecycle.html

image

Correct url is just https://lf-lynx.tiktok-cdns.com/obj/lynx-artifacts-oss-sg/lynx-website/assets/lifecycle-init-render-1.png


rayfarer:
I also noticed the zh version had the same problem, so added that to the PR

PR #24, opened 2025-03-08
Updated QRCodeSVG size from default 128 to 220 for easy scan for qrcode scanner

NzakiCodes:
Updated QRCodeSVG size from default 128 to 220 for easy scan for QRCode scanners on android devices.

link: https://lynxjs.org/guide/start/quick-start.html#ios-simulator-platform=macos-arm64,explorer-platform=android

Change-Id: I0e1f3e6aedabf90fd1b5c753264c7e647e03f733


colinaaa (approved):
Thank you!

PR #31, opened 2025-03-10
feat: update link

Aimee1608:
Change-Id: I7094695e70f0fc80b659bac41fbeae70531e63e6

### DIFF
--- a/src/components/go/example-preview/components/index.tsx
+++ b/src/components/go/example-preview/components/index.tsx
@@ -278,9 +278,10 @@ export const ExampleContent: FC<ExampleContentProps> = ({
                       {t('go.scan.message-1')}
                       <Typography.Text
                         link={{
-                          href: lang
-                            ? `/${lang}/guide/start/quick-start.html`
-                            : '/guide/start/quick-start.html',
+                          href:
+                            lang === 'zh'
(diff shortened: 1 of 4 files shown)


src/components/go/example-preview/components/index.tsx
```diff
@@ -278,9 +278,10 @@ export const ExampleContent: FC<ExampleContentProps> = ({
...
                       <Typography.Text
                         link={{
-                          href: lang
-                            ? `/${lang}/guide/start/quick-start.html`
-                            : '/guide/start/quick-start.html',
+                          href:
+                            lang === 'zh'
+                              ? `/${lang}/guide/start/quick-start.html#download-lynx-explorer,ios-simulator-platform=macos-arm64,explorer-platform=ios-simulator`
+                              : '/guide/start/quick-start.html#download-lynx-explorer,ios-simulator-platform=macos-arm64,explorer-platform=ios-simulator',
                           target: '_blank',
```
jianliang00, line 285: When clicking the link, the same page opens in a new tab. Would it be smoother if we removed the target="_blank" attribute?

Aimee1608: There are many examples on this page.such as https://deploy-preview-31--lynx-doc.netlify.app/guide/ui/layout/linear-layout.html . If you don't open them in a new tab, after navigating away, you can't return to the specific example you were focusing on previously.

PR #38, opened 2025-03-10
fix wrong code in integrating-lynx-with-web

Sodiumsss:
According to the above steps, the URL here should be written as url="/main.web.bundle" instead of url="/main.lynx.bundle". If url="/main.lynx.bundle" is used, the page will not display correctly.


jonathanfilbert (approved):

colinaaa (approved):
Thank you!

PR #42, opened 2025-03-11
docs(en): Add nanostores to state manager list

wobsoriano:


Huxpro:
@wobsoriano Thanks for bringing up nanostore to our attention!

While we may not merge this PR as-is for now, we will:

1. Take this opportunity to consider how we can better showcase community packages
2. Bring this back to our internal team to see if we can test and adopt it (since we typically only feature tools we actively use)

Thanks for the understanding and stay tuned!

wobsoriano:
Awesome! Thanks for checking it still 🫡

PR #52, opened 2025-03-11
Use the default bundle name for web

dobladov:
Initially I tried to run the bundle as is presented in the docs `main.lynx.bundle`, it's much clearer to use `main.web.bundle` so there are no confusion to which bundle to use.

close: #55


babu-ch:
ref
https://github.com/lynx-family/lynx-website/issues/55

colinaaa (approved):
Thank you!

Issue #23, opened 2025-03-08
RSS support for the blog


kuchta:
Please implement RSS support for the blog so we can follow 🙏🏻

SoonIter:
Rspress supports RSS natively

https://rspress.dev/plugin/official-plugins/rss#rspressplugin-rss

So, pr welcome :D 

#### Related info

https://github.com/web-infra-dev/rspack/blob/c8bcb8ffc0504227123bc66b62529bc63f971cf2/website/rspress.config.ts#L41

https://x.com/Soon_Iter/status/1845816148191048027

babu-ch:
@SoonIter 
hi, I also want to follow the information so I created a PR. 😄

Huxpro:
Hi @SoonIter could you do a review over the PR? I just invited you to the repo.

babu-ch:
It has been released and can CLOSE!

PR #75, opened 2025-03-14
feat: rss

babu-ch:
resovle https://github.com/lynx-family/lynx-website/issues/23

Almost the same as this setting.
https://github.com/web-infra-dev/rspack/blob/c8bcb8ffc0504227123bc66b62529bc63f971cf2/website/rspress.config.ts#L41

Change-Id: I2756aff45971785c9a661776912380e4ad73d777

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@mdn/minimalist": "^2.0.4",
     "@rsbuild/plugin-svgr": "^1.0.7",
     "@rspress/plugin-auto-nav-sidebar": "1.43.0-canary-202503021900",
+    "@rspress/plugin-rss": "^1.43.4",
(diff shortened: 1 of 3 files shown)


SoonIter (approved):
LGTM❤️️

image

package.json
```diff
@@ -56,6 +56,7 @@
     "@mdn/minimalist": "^2.0.4",
     "@rsbuild/plugin-svgr": "^1.0.7",
     "@rspress/plugin-auto-nav-sidebar": "1.43.0-canary-202503021900",
+    "@rspress/plugin-rss": "^1.43.4",
```
SoonIter, line 59: the versions of "@rspress/*" should be the same

But it does not matter, I'll bump the lynx website to 2.0 alpha

PR #80, opened 2025-03-16
docs: update links in scrolling guide to English version

kickbelldev:


Huxpro (commented):
Thanks for working on this. Ideally...neither the en docs or zh docs should hardcode the language prefix as Rspress should be able to handle it correctly.

Could you help with removing both?

kickbelldev:
Thanks for the review! I've removed the hardcoded language prefixes (`en/` and `zh/`) from all internal documentation links.

Changes made:
- Removed `en/` prefix from English docs
- Removed `zh/` prefix from Chinese docs
- Converted relative paths to absolute paths for better maintainability

Huxpro (approved):
Thanks a lot!

PR #90, opened 2025-03-19
fix: enhance type safety in lynx-compat-data

kickbelldev:
Change-Id: I87171c08bbc1589f9522ef311887ee1e4f08566b


Huxpro:
@kickbelldev thanks for working on this! 

While i'd love to merge this, @minalwws @colinaaa It doesn't seem like we run any `lynx-compat-data` test from the Github CI?

PR #100, opened 2025-03-24
opt: Deleted documents related to the TTI metric as it has been depre…

Tamerlx:
…cated.


partholon:
> still need to remove "Main thread functions can only execute after TTI (Time to Interactive). This means they cannot execute during the initial screen load." in react/Document.directives and react/main-thread-script

No need to modify, in those two articles, TTI is used to refer to a certain point in time, not the indicator itself

others LGTM.

partholon (approved):

PR #105, opened 2025-03-26
Adjust the image layout in the element panel and layer panel

yuweizheng:
Change-Id: I56e58d04874a23c9e132d034590036e26ba110c2


talisk (approved):
All changes are updated imgs. Very large workload, thank you for your hard work.

PR #114, opened 2025-03-31
State Variable Fix (shouldShow to shouldDisplay)

kelvinelove:
There was a typo in the conditional rendering of the lazyload component. We don't have a component called `shouldShow`. It's supposed to be `shouldDisplay`


colinaaa (approved):
Thank you!

PR #117, opened 2025-04-07
feat: add guide and api docs for react lynx testing library

upupming:
In https://github.com/lynx-family/lynx-stack/pull/74, we implemented `@lynx-js/lynx-environment` and `@lynx-js/react-lynx-testing-library` for ReactLynx Testing Library.

This PR includes guides and API docs for those two packages.


docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,404 @@
...
+        <text>
+          hey
+        </text>
+      </page>
+    `);
+  }
+});
+```
+
+### More usage
```
hzy, line 616: `<list/>` will be in "More usage", or should we guide here?

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,404 @@
...
+    // ...
+  },
+});
+
+export default defineConfig(mergeConfig(defaultConfig, config));
+```
+
+Then you can start writing tests and running them!
+
+:::info No need to install @lynx-js/environment
```
hzy: I personally feel a little bit strange reading here...

Do you want to emphasize "no need to install" or "although we are based on @lynx-js/environment, we have already packaged it for you"? If it is the latter, you can wrap this part with `<Details/>`?

```tsx
import { Details } from '@lynx';

<Details>
...
</Details>
```

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,404 @@
...
+```
+
+### Rerender
+
+```jsx
+import '@testing-library/jest-dom';
+import { render } from '@lynx-js/react-lynx-testing-library';
+import { expect } from 'vitest';
+
+it('rerender will re-render the element', async () => {
```
hzy: nit
```suggestion
it('rerender will re-render your component', async () => {
```

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,404 @@
...
+      <text>
+        hi
+      </text>
+    </page>
+  `);
+  expect(container.firstChild).toHaveTextContent('hi');
+
+  {
+    // Unlike React Testing Library, container needs to be retrieved after rerender
+    // Because ReactLynx creates a new page element each time it loads
```
hzy, line 447: I can't quite remember but we created a module-scoped persistent page and stored it in a `__page` variable?

docs/en/react/react-lynx-testing-library/index.mdx
upupming: I have added the list and mts example, PTAL again~

Huxpro:
@minalwws @colinaaa Does anyone know why the Netlify CI doesn't pick up this PR and build us a preview?

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,624 @@
...
+  expect(container).toMatchInlineSnapshot(`
+    <page>
+      <list
+        update-list-info="[{"insertAction":[{"position":0,"type":"__Card__:__snapshot_f75b7_test_2","item-key":0},{"position":1,"type":"__Card__:__snapshot_f75b7_test_2","item-key":1},{"position":2,"type":"__Card__:__snapshot_f75b7_test_2","item-key":2}],"removeAction":[],"updateAction":[]}]"
+      />
+    </page>
+  `);
+  const list = container.firstChild;
+  // Trigger the componentAtIndex method of list and load the 0th item
+  const uid0 = elementTree.triggerComponentAtIndex(list, 0);
```
hzy: Before we bring these internal api into public, can we re-consider the naming of it?

At least replace `Component` with `listItem` or something similar :D

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,624 @@
...
+        },
+      ],
+    ]
+  `);
+});
+```
+
+#### Choose the thread used for rendering
+
+The second parameter of the `render` method supports `enableMainThread` and `enableBackgroundThread` are two options for enabling the main thread and background threads. By default, `enableMainThread` is `false` and `enableBackgroundThread` is `true`, so the main thread does not render the first screen, only the background thread runs the full Preact runtime and sends UI updates to the main thread for rendering. The reason why we don't enable both the main thread and the background thread is because unit tests do not compile files into two products like using Rspeedy does, which will result in the top-level code being executed only once, and by default in the background thread. Suppose you wrote a component like this:
```
hzy: nit:
```suggestion
The second parameter of the `render` method supports `enableMainThread` and `enableBackgroundThread` are two options for enabling the main thread and background threads. By default, `enableMainThread` is `false` and `enableBackgroundThread` is `true`, so the main thread does not render the first screen, only the background thread runs the full Preact runtime and sends UI updates to the main thread for rendering. The reason why we don't enable both the main thread and the background thread is because unit tests do not compile files into two bundles like using Rspeedy does, which will result in the top-level code being executed only once, and by default in the background thread. Suppose you wrote a component like this:
```

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,624 @@
...
+const Comp = () => {
+  return (
+    <view>
+      <text>{isBackground ? 'background' : 'main thread'}</text>
+    </view>
+  );
+};
+```
+
+The top-level `isBackground` of this component will only be executed once, and by default it is `true` (Lynx Environment defaults to the background thread after initialization). In order to avoid rendering the wrong first screen result when `enableMainThread: true`, we set `enableMainThread` to `false` by default, and get an empty first screen, which can avoid this problem.
```
hzy: nit:
```suggestion
The top-level `isBackground` of this module will only be executed once, and by default it is `true` (Lynx Environment defaults to the background thread after initialization). In order to avoid rendering the wrong first screen result when `enableMainThread: true`, we set `enableMainThread` to `false` by default, and get an empty first screen, which can avoid this problem.
```

docs/en/react/react-lynx-testing-library/index.mdx
upupming: - `triggerComponentAtIndex` calls `componentAtIndex` of list element
- `triggerEnqueueComponent` calls `enqueueComponent` of list element

However, `componentAtIndex` and `enqueueComponent` are not directly exposed to users.

Let me use `enterListItemAtIndex` and `leaveListItem` for now.

https://github.com/upupming/lynx-stack/blob/97557eb25c3466a6fba8a28713155fa621e9de4f/packages/testing-library/lynx-environment/src/lynx/ElementPAPI.ts#L437

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,622 @@
...
+  const { container } = render(<Comp />);
+  expect(container).toMatchInlineSnapshot(`
+    <page>
+      <list
+        update-list-info="[{"insertAction":[{"position":0,"type":"__Card__:__snapshot_f75b7_test_2","item-key":0},{"position":1,"type":"__Card__:__snapshot_f75b7_test_2","item-key":1},{"position":2,"type":"__Card__:__snapshot_f75b7_test_2","item-key":2}],"removeAction":[],"updateAction":[]}]"
+      />
+    </page>
+  `);
+  const list = container.firstChild;
+  // Trigger the componentAtIndex method of list and load the 0th item
```
hzy: nit: The comment is inconsistent with code below

docs/en/react/react-lynx-testing-library/index.mdx
```diff
@@ -0,0 +1,622 @@
...
+      <list-item
+        item-key="0"
+      >
+        <text>
+          0
+        </text>
+      </list-item>
+    </list>
+  `);
+  // Trigger the enqueueComponent method of list to notify that 0th item is recycled
```
hzy: nit: The comment is inconsistent with code below

hzy (approved):

PR #119, opened 2025-04-10
docs: fix example code 

hackmajoris:


Huxpro:
@hackmajoris Thanks! Could you also help fix the Chinese version?

hackmajoris:
> @hackmajoris Thanks! Could you also help fix the Chinese version?

Done!

PR #152, opened 2025-05-08
feat: upgrade Rspress-2.0.0-beta.4 and add

SoonIter:
Change-Id: Ifa2b96a6a7ab68192ac666562e4e08e8ef651325


colinaaa (approved):
LGTM 💯

docs/en/_meta.json
```diff
@@ -1,21 +1,21 @@
 [
   {
-    "text": "nav.guide",
+    "text": "Guide",
```
Huxpro, line 3: IIRC `nav.guide` was declared in `i18n.json`, why do we want to replace it with plain text?

Huxpro (approved):
Awww! LET'S GO!

Huxpro:
> dynamic TOC support

I'd love to know more about how you did it and if there is any constraints on how MDX (and MDX components) needs to be written ;)

docs/en/_meta.json
SoonIter: @rspress/plugin-llms generate the llms.txt based on `_meta.json`, but it does not read the `i18n.json` now...

so this is a workaround, it will be supported later

image

SoonIter:
> > dynamic TOC support
> 
> I'd love to know more about how you did it and if there is any constraints on how MDX (and MDX components) needs to be written ;)

@Huxpro 

I use `MutationObserver` to watch the h2-h4 dom change to trigger toc update,  `.rspress-doc > h2-h4`. The usage of the mdx component has not changed at all. 

But I just added a single layer of listening due to performance reasons, so I add this hack

https://github.com/lynx-family/lynx-website/blob/6436d15726770443b77cb43eaabd66c71721ae87/src/components/platform-tabs/PlatformTabs.tsx#L274-L276

PR #175, opened 2025-05-19
Add an example of extending the `GlobalProps` interface

plesiecki:
Hello, I'd like to provide an example of how to extend the `GlobalProps` interface to make TS "happy" and save other users the trouble.

### DIFF
--- a/docs/en/api/lynx-api/lynx/lynx-global-props.mdx
+++ b/docs/en/api/lynx-api/lynx/lynx-global-props.mdx
@@ -8,7 +8,15 @@
 
 ## Usage
 
-```jsx
+```tsx
+// Properties inside `lynx.__globalProps` is NOT managed by Lynx, so you MUST extends this interface yourself
+declare module '@lynx-js/types' {
+  interface GlobalProps {
+    appTheme: string;
+    title: string;
+  }
+}
+
 function App() {
   const themeClass = useMemo(
     () => `theme-${lynx.__globalProps.appTheme}`,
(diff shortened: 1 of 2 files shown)


docs/en/api/lynx-api/lynx/lynx-global-props.mdx
```diff
@@ -8,7 +8,14 @@
 
 ## Usage
 
-```jsx
+```tsx
+declare module '@lynx-js/types' {
```
Huxpro, line 13: Could you pre-append a sentence of comment? (Just ChatGPT the Chinese version)

Huxpro (commented):
Hii @plesiecki thanks for adding this. This will be helpful

docs/en/api/lynx-api/lynx/lynx-global-props.mdx
plesiecki: Inside the code block or outside?

colinaaa: I believe it makes more sense to include it internally, as comments are just a part of the code.

plesiecki: I've used your own words from https://github.com/lynx-family/lynx/blob/develop/js_libraries/types/types/background-thread/lynx.d.ts#L80. Hopefully the ChatGPT translation is accurate.

colinaaa (approved):
Thank you!

PR #182, opened 2025-05-22
Improve spelling and wording

plesiecki:
Some spelling errors slipped unnoticed last time. Here is an improved version.


Huxpro:
@plesiecki perfect.

I do think we need to document more knowledges regarding how to better use of types so I created an issue for us to track: https://github.com/lynx-family/lynx-website/issues/183

Let us know when you are running into type related issues or gotchas that are worth documenting!

PR #185, opened 2025-05-23
Add setup for product detail page

f0rdream:
Add setup for product detail page, and add pics.ts into `<Go>`'s entries.


f0rdream:
@Huxpro Have a look

PR #197, opened 2025-05-30
doc: modify list main/cross axis gap description

DwwWxx:
Change-Id: I448cae122a35c1640ba9864e530c5814bb85de74

### DIFF
--- a/docs/en/api/elements/built-in/list.mdx
+++ b/docs/en/api/elements/built-in/list.mdx
@@ -52,6 +52,40 @@ Waterfall Layout Example:
   highlight="{14-15}"
 />
 
+## List-specific CSS Properties
+
+:::note
+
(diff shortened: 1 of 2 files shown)


docs/en/api/elements/built-in/list.mdx
```diff
@@ -52,6 +52,34 @@ Waterfall Layout Example:
   highlight="{14-15}"
 />
 
+## CSS Attributes
```
Huxpro:
```suggestion
## List-specific CSS Properties

:::note What are List-specific CSS properties?
(explain that such CSS properties are only applicable on `<list>` and will have no effects otherwise)
:::
```

Huxpro (approved):
Please fix the nits before merging

docs/en/api/elements/built-in/list.mdx
Huxpro: Don't forget about the Chinese

Huxpro (approved):

Issue #183, opened 2025-05-22
Guide on how to better use TypeScript with Lynx


Huxpro:
There are many undocumented knowledges regarding how to enforce type safety when developing the Lynx apps, e.g.

1. how to configure JSX properly: https://github.com/lynx-family/lynx-stack/pull/545
2. how to extend untyped parameters. e.g. [`GlobalProps`](https://github.com/lynx-family/lynx-website/pull/175)
3. how to extend types for native modules and custom native elements. 

We need to introduce a new guide around https://github.com/lynx-family/lynx-website/issues/176

PR #200, opened 2025-06-03
docs: extending `NativeModules`

colinaaa:
issue: https://github.com/lynx-family/lynx-website/issues/183.

You may also see an example in https://github.com/lynx-family/lynx-examples/pull/78.

> [!NOTE]
> I didn't use `<Go />` since there is not TypeScript intellisense there :)


Huxpro (approved):
lgtm

PR #272, opened 2025-07-30
fix(types): fix some type errors

SoonIter:
just small fix of type errors

pre-tasks for https://github.com/lynx-family/lynx-website/pull/251


colinaaa (approved):
👍

PR #332, opened 2025-08-28
feat: add pnpm dlx upgrade rspeedy

zoolsher:
Change-Id: Ie410f80f0ab8011392b2606789eabd201e2bbb3e


docs/en/rspeedy/upgrade.md
```diff
@@ -22,6 +22,12 @@ To upgrade `@lynx-js/rspeedy` and its plugins to the latest version, use the fol
 npx upgrade-rspeedy@latest
```
colinaaa: Can we use `<PackageManagerTabs>`, instead?

zoolsher: sure

colinaaa: It looks like there is an `exec` option: https://v2.rspress.rs/guide/default-theme/components#packagemanagertabs

```ts
      /**
       * If true, the command will be interpreted as a shell command and prefixed with npx for npm,
       * or the package manager binary for others.
       */
      exec?: boolean;
    }
```

PR #339, opened 2025-09-01
chore(web): caution global-bind not supported in web

Sherry-hue:
Change-Id: I14ad06fbf6aa88471a135d899784bcdfacd04304


docs/en/guide/interaction/event-handling/event-propagation.mdx
```diff
@@ -90,6 +90,11 @@ In the above example, since the `click me` area sets the static interception `ta
 
 ## Cross-component event listening
 
+:::caution
+Android and iOS only.
```
colinaaa: This is not correct. We also have Harmony (and other platforms in the future).

docs/en/guide/interaction/event-handling/event-propagation.mdx
```diff
@@ -90,6 +90,11 @@ In the above example, since the `click me` area sets the static interception `ta
...
+There is no support for Web platform.
```
colinaaa: Can we use compat-table instead?

Huxpro: consider using APIBadge `<NoWeb>`(<https://github.com/lynx-family/lynx-website/blob/main/src/components/api-badge/index.ts>) to indicate compatibility in guide 


E.g. <https://lynxjs.org/api/elements/built-in/view.html#flatten>

image

PR #359, opened 2025-09-08
feat: add `create` command to homepage

colinaaa:
Also, allow overriding the HomeLayout by props.


theme/index.tsx
```diff
@@ -164,18 +165,43 @@ function HomeLayout() {
...
+
+  // Rspress would pass `afterHero: undefined` and `afterHeroActions: undefined` props to HomeLayout,
+  const {
+    afterHero = (
+      <>
+        <Features src={routePath} /> {routePath === '/' && <ShowCase />}
+        {routePath === '/' && <Banner />}
+      </>
+    ),
+    afterHeroActions = (
```
zoolsher, line 178: Image

Add 50% transparency to make this block blend better with the background.

colinaaa: image

PR #364, opened 2025-09-09
docs: add Router configuration to Tanstack Router

gaoachao:
follow up https://github.com/lynx-family/lynx-examples/pull/161

Change-Id: I0a171045e32538b7fe07bf47cfc24cd5373b2f15

### DIFF
--- a/docs/en/react/routing/tanstack-router.mdx
+++ b/docs/en/react/routing/tanstack-router.mdx
@@ -79,7 +79,7 @@ const memoryHistory = createMemoryHistory({
 const router = createRouter({
   routeTree,
   history: memoryHistory,
-  isServer: false,
+  isServer: false, // [!code highlight]
 });
 
 export function App() {
--- a/packages/lynx-example-packages/package.json
+++ b/packages/lynx-example-packages/package.json
@@ -39,7 +39,7 @@
     "@lynx-example/swiper": "0.4.0",
     "@lynx-example/text": "0.5.2",
     "@lynx-example/view": "0.5.0",
-    "@lynx-example/tanstack-router": "0.6.2",
+    "@lynx-example/tanstack-router": "0.6.4",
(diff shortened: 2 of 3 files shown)


docs/zh/react/routing/tanstack-router.mdx
```diff
@@ -55,6 +55,36 @@ export default defineConfig({
...
+import { routeTree } from './routeTree.gen.js';
+
+const memoryHistory = createMemoryHistory({
+  initialEntries: ['/'],
+});
+
+const router = createRouter({
+  routeTree,
+  history: memoryHistory,
+  isServer: false,
```
colinaaa: We could highlight this

```suggestion
  isServer: false, // [!code highlight]
```

docs/zh/react/routing/tanstack-router.mdx
```diff
@@ -55,6 +55,36 @@ export default defineConfig({
...
+});
+
+export function App() {
+  return <RouterProvider router={router} />;
+}
+```
+
 ### Polyfill
 
 为了确保 [URLSearchParams API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) 在 Lynx 环境中正常工作，请在入口文件顶部添加 `url-search-params-polyfill`：
```
colinaaa, line 90: Instead of `import 'url-search-params-polyfill';`, can we use `source.preEntry` instead?

colinaaa (commented):
Should we also update the example in `<Go />`?

docs/zh/react/routing/tanstack-router.mdx
gaoachao: I think `source.preEntry` can also work, but it requires having a `url-search-params-polyfill.js` file under the `src` or project root directory, which has no advantage compared to directly using `import 'url-search-params-polyfill';`.

PR #372, opened 2025-09-15
docs: add __GetComputedStyleByKey PAPI document

usczz:
Change-Id: I26abbefa5741fd4dc79283f5a81c1561edadd2a7

### DIFF
--- a/docs/en/api/engine/element-api/__GetComputedStyleByKey.mdx
+++ b/docs/en/api/engine/element-api/__GetComputedStyleByKey.mdx
@@ -1,0 +1,54 @@
+# `__GetComputedStyleByKey`
+
+## Introduction
+
+Returns the resolved style value for a CSS property of an element
+
+## Syntax
+
+```jsx
+__GetComputedStyleByKey(element: Element, key: string) : string;
+```
+
+### Parameters
+
+|   Key   | Description                                                                                                                                                                              |
+| :-----: | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| element | Any Element object.                                                                                                                                                                      |
+|   key   | A string indicating the CSS property key to be retrieved. Keep in mind that not every CSS property is supported. Refer to the Compatibility table to see which properties are supported. |
+
+### Return Value
+
+Resolved style value for a CSS property of specified element.
+
+:::caution
+Unlike browser, this PAPI won't trigger document reflow.
+:::
+
+## Example
+
+The frontend framework can retrieve resolved style value of an Element as follows.
+
+```jsx
+// main-thread.js
+let page = __CreateElement('page', 0, {});
+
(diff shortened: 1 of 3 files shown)


packages/lynx-compat-data/elements/get-computed-style-supported-properties.json
```diff
@@ -0,0 +1,248 @@
+{
+  "elements": {
+    "get-computed-style-supported-properties": {
+      "top": {
+        "__compat": {
+          "description": "top",
+          "status": {
+            "experimental": true
```
colinaaa, line 8: Is this API still experimental in Lynx 3.5?

docs/en/api/engine/element-api/__GetComputedStyleByKey.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+## Example
+
+The frontend framework can retrieve resolved style value of an Element as follows.
+
+```jsx
+// main-thread.js
+let element = __CreateElement('view', 0, {});
+
+// after connecting to DOM tree and trigger reflow
```
colinaaa: Instead of this description, can we use a code example? E.g.:

```js
let parent = __CreateElement('view', 0, {})
let child = __CreateElement('view', 0, {})
__AppendElement(parent, child)

let width = __GetComputedStyleByKey(child, 'width');
```

docs/en/api/engine/element-api/__GetComputedStyleByKey.mdx
```diff
@@ -0,0 +1,44 @@
...
+```jsx
+__GetComputedStyleByKey(element: Element, key: string) : string;
+```
+
+### Parameters
+
+|   Key   | Description                                                 |
+| :-----: | ----------------------------------------------------------- |
+| element | Any Element object.                                         |
+|   key   | A string representing the CSS Property key to be retrieved. |
```
colinaaa: Can we also make it clear that not all CSS properties are supported now

packages/lynx-compat-data/elements/get-computed-style-supported-properties.json
usczz: Yes, this PAPI is experimental in Lynx 3.5 for all supported properties. Its behavior may change in later SDK version.

docs/en/api/engine/element-api/__GetComputedStyleByKey.mdx
usczz: Good idea, replaced with more straightforward example.

docs/en/api/engine/element-api/__GetComputedStyleByKey.mdx
usczz: Good idea, explicitly add supporting status for other CSS properties in compatibility table.

PR #431, opened 2025-10-02
Mention `create-lynxjs-app` CLI in Quick Start doc

szymonrybczak:
Change-Id: Iadfb0938f36b1e5190d4c40ad6ea230cb4207378

### DIFF
--- a/.cspell/dictionary.txt
+++ b/.cspell/dictionary.txt
@@ -28,6 +28,7 @@ jsbridge
 jscpuprofile
 jssdk
 lynxdevtool
+lynxjs
 lynxsdk
 mousedbclick
 napi


docs/en/guide/start/quick-start.mdx
```diff
@@ -14,6 +14,21 @@ Welcome to the Lynx documentation! We will create a Lynx project and start devel
 
 ## Installation
 
+### Create a new Lynx app
```
Huxpro, line 17: Since the current default path is brownfield and we are now introducing a new greenfield path, I suggest we described that we have two paths clearly upfront at the start of Quick Start (somewhat similar to how React introduces "From the scratch" and "Framework"):

image

And then we can link to a sub-section within this page titled something like "Create a Lynx project with the Native App Template"

PR #444, opened 2025-10-15
feat: mention lazy components css is scoped

upupming:
Change-Id: I7fc7c5a6a2a8fa332d703088c4fc1522b396850f

Related Discussion: https://github.com/lynx-family/lynx/discussions/3160

### DIFF
--- a/docs/en/react/code-splitting.md
+++ b/docs/en/react/code-splitting.md
@@ -87,6 +87,10 @@ export function App() {
 }
 ```
 
+:::warning Differences from the Web
+In Lynx, CSS is scoped to each [Bundle](../api/lynx-native-api/template-bundle.md). Since lazy-loaded components generate their own Bundles (see output structure at [Output Files](../rspeedy/output.md)), you shouldn’t expect them to behave like on the Web. Global CSS defined in the main Bundle will not affect lazy-loaded components, and vice versa.
+:::
+
 ### Error handling
 
 #### Use ErrorBoundary
--- a/docs/zh/react/code-splitting.md
+++ b/docs/zh/react/code-splitting.md
@@ -83,6 +83,10 @@ export function App() {
 }
 ```
 
+:::warning 与 Web 的区别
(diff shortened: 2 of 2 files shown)


docs/zh/react/code-splitting.md
```diff
@@ -83,6 +83,10 @@ export function App() {
 }
 ```
 
+:::info
+懒加载组件会生成独立的 Bundle，其 CSS 和主 Bundle 相互隔离。主 Bundle 定义的全局 CSS 不会在懒加载组件生效，反之亦然。
+:::
```
Huxpro: 1. This should be a warning titled "与 Web 的区别"

2. Link "Bundle" to other doc (guide or spec), otherwise this concept is undefined 

3. We should probably use words defined by CSS specification and link to it. Something like "In Lynx, CSS are scoped/layered according to each Bundle, and since lazy-loaded components are generating its own Bundle...you may not expect it to work like Web"

Cc @zhongyr for second opinion

Huxpro (changes requested):

docs/zh/react/code-splitting.md
upupming: - 1 and 2 are updated.
- For 3, We do not officially define "scoped" in our Docs. So I mentioned the config `enableRemoveCSSScope` and tell user not to confuse scope created by `enableRemoveCSSScope` and scope in different bundles.

Huxpro (approved):
There is one nit but this is great!

docs/en/react/code-splitting.md
```diff
@@ -87,6 +87,12 @@ export function App() {
 }
 ```
 
+:::warning Differences from the Web
+In Lynx, CSS is scoped to each [Bundle](../api/lynx-native-api/template-bundle.md). Since lazy components generate their own Bundles (see output structure at [Output Files](../rspeedy/output.md)), you shouldn’t expect them to behave like on the Web. Global CSS defined in the main Bundle will not affect lazy components, and vice versa.
```
Huxpro: Ideally speaking, there should be only "lazy-loaded components" and "`lazy ` component". 

You may want to update this and other occurrences of "lazy components" in this page

upupming: Updated. Thanks! (Other occurrences in this page are all lazy-loaded components, no need to change.)

PR #465, opened 2025-10-21
P/ljl/add version page

loongliu:


SoonIter (approved):

docs/en/versions.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+## Next version (unreleased)
+
+Developers can check out the new features currently in development in this release.
+
+| **version number** | **docs repo branch**                                          | **docs link**           |
+| ------------------ | ------------------------------------------------------------- | ----------------------- |
+| next               | [main](https://github.com/lynx-family/lynx-website/tree/main) | https://lynxjs.org/next |
+
+## Release candidate version
```
Huxpro, line 17: This sounds cool but I don’t think we have a defined process yet for "releasing RC" and "moving an RC to an stable release".

docs/public/version.json
```diff
@@ -0,0 +1 @@
+{ "versions": ["next", "3.5", "3.4"] }
```
Huxpro, line 1: Oh ok so you defined `version.json` in this PR... it's probably easier to review if these were one single PR

docs/en/versions.mdx
```diff
@@ -0,0 +1,38 @@
+---
+sidebar: false
+---
+
+# Lynx versions
```
Huxpro, line 5: Since we already define versions in `versions.json`, it might actually be better in a long run to define everything as metadata in that structured json and render it as markdown instead of maintaining raw markdown here so there is a single source of truth to update for every releases

E.g.

```json
{
  "version": [
    {
      "3.4": {
        "gh_tag": "3.4.1",
        "blog": "..."
      }
    },
  ],
}
```

PR #545, opened 2025-11-26
docs(xelement): update generated docs

MoonfaceX:
Automated XElement Doc Generation update


Huxpro:
Closing as stale generated output. These docs should be regenerated from the current generator and source data rather than landing an old snapshot.

PR #896, opened 2026-04-08
fix(docs): auto fixing by `pnpm run fix-docs`

Dugyu:
Run `fix-docs` script and commit all changes

Change-Id: I21ce26bdda17fa0cf6118fb938734f06e6e1169d

### DIFF
--- a/docs/en/api/elements/built-in/scroll-view-API.mdx
+++ b/docs/en/api/elements/built-in/scroll-view-API.mdx
@@ -1,11 +1,23 @@
-import { AndroidOnly, IOSOnly, ClayOnly, ClayWindowsOnly, ClayMacOSOnly, HarmonyOnly, Required, Deprecated, Go, PlatformTabs } from '@lynx';
-import { VersionBadge, APITable } from '@lynx';
+import {
+  APITable,
+  AndroidOnly,
+  ClayMacOSOnly,
+  ClayOnly,
+  ClayWindowsOnly,
+  Deprecated,
+  Go,
+  HarmonyOnly,
+  IOSOnly,
+  PlatformTabs,
+  Required,
+  VersionBadge,
+} from '@lynx';
 
 ## Attributes
 
 ### `bounces`
 
- <IOSOnly /> <ClayOnly /> <HarmonyOnly /> <VersionBadge>1.4</VersionBadge>
+<IOSOnly /> <ClayOnly /> <HarmonyOnly /> <VersionBadge>1.4</VersionBadge>
 
 ```tsx
 // @defaultValue: true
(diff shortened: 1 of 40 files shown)


docs/en/api/elements/built-in/scroll-view-API.mdx
```diff
@@ -98,102 +109,102 @@ Replacement of scroll-x and scroll-y
...
-    .exec();
-
-```
+.select('#id')
+.invoke({
+method: 'autoScroll',
+params: {
+/**
+_ The distance of each second's scrolling, which supports positive and negative values. The unit of distance can be "px", "rpx", "ppx", or null (for iOS, the value must be greater than 1/screen.scale px).
+_ @Android
```
MoonfaceX, line 184: PLS DO NOT FORMAT AUTO-GENERATED MDs

Huxpro (commented):
Thanks for running `fix-docs` to clean up formatting. Most of the changes here look like standard auto-formatting (import sorting, whitespace normalization, markdown table alignment), which is great.

However, I spotted a few categories of **potentially breaking changes** that appear to be introduced by the formatter rather than being intentional fixes. Please double-check these before merging:

### 1. Code block indentation stripped inside fenced blocks

In several files (e.g., `image-API.mdx`, `input-API.mdx`, `scroll-view-API.mdx`, and their `zh` counterparts), the formatter removed all indentation from JS code inside ` ```ts ` blocks. For example, the `pauseAnimation` method call chain went from properly indented:

```js
lynx.createSelectorQuery()
     .select('#id')
     .invoke({
      method: 'pauseAnimation',
      ...
    })
    .exec();
```

to completely flat:

```js
lynx.createSelectorQuery()
.select('#id')
.invoke({
method: 'pauseAnimation',
...
})
.exec();
```

This makes the code examples significantly harder to read for users.

### 2. Triple backticks changed to quadruple backticks

Closing ` ``` ` markers for code blocks were changed to ` ```` ` (four backticks) in many places. While this might still render correctly in some markdown parsers, it's inconsistent with the opening markers which remain triple backticks. This could cause rendering issues.

### 3. JSDoc comments corrupted in TypeScript code blocks

In `input-API.mdx` and `textarea-API.mdx` (both `en` and `zh`), JSDoc comments inside code blocks were corrupted by the formatter — `*` was replaced with `_`, and `*/` became `\*/`. For example:

```
- * @Web
- * @since 3.4
- */
+ _ @Web
+ _ @since 3.4
+ \*/
```

This breaks the JSDoc syntax and would render incorrectly.

### 4. Unintended escaping of `*` in prose

In `image-API.mdx`, a multiplication sign in prose was escaped:
```
- (1024 * 768 * bits per pixel / 8)
+ (1024 * 768 \* bits per pixel / 8)
```

This would render as a literal backslash in the docs.

---

**Recommendation**: It seems like the `fix-docs` script may have some overly aggressive rules or bugs that are causing these issues. I'd suggest:
1. Verifying the rendered output for the affected pages (especially the code examples)
2. If these are indeed bugs in the formatter, fixing the formatter config and re-running
3. At minimum, the JSDoc corruption (issue 3) and code indentation stripping (issue 1) should be fixed before merging

PR #957, opened 2026-04-29
docs: updates some assets url host for all user

toretto-wt:
Change-Id: Ieb3b0bae92dc22f21f2145e1dd5748515a052f4e

### DIFF
--- a/docs/en/api/elements/built-in/frame.mdx
+++ b/docs/en/api/elements/built-in/frame.mdx
@@ -27,7 +27,7 @@ The following is an example of a main page loading an embedded page via the `<fr
 ### Main Page
 
 <Go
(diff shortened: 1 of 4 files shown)


Huxpro (dismissed):
Thanks for moving these assets to the new CDN host. I checked the changed URLs and the intended `https://lf-lynx.tiktok-cdns.com/...` assets return `200` with the expected image content types, and the Netlify preview/checks are green.

One blocking issue before merge:

- `docs/en/api/elements/built-in/frame.mdx` changes the main page image URL to `hhttps://lf-lynx.tiktok-cdns.com/.../frame-example.jpg` — note the extra leading `h`. That will produce a broken image in the English `<frame>` docs.

Please fix that typo to `https://...`. After that, this looks safe to approve.

MoonfaceX (approved):

PR #968, opened 2026-05-06
feat: update lynx-ui contributing

f0rdream:
Change-Id: I84c75cdf170f6a699025c1a6f671b79c69c53112


Huxpro (commented):
Good cleanup overall — the rewrite is much more focused and appropriately defers the full workflow to the upstream CONTRIBUTING.md. Deleting the WIP placeholders and fixing the `_meta.json` link from `/api/lynx-api/main-thread.html` to `/react/main-thread-script.html` are both correct.

Two items to check:

1. Inconsistent script name — In both EN and ZH contributing docs, the "create" section uses `pnpm make-new-component --create` (kebab-case) while the "example" section uses `pnpm makeNewComponent --example` (camelCase). These should match whichever npm script name actually exists in the lynx-ui repo. If the script was renamed, the example command needs the same update.
2. Node.js >= 24.0.0 — The old docs specified >= 20. Node 24 was released April 2025 and won't reach LTS until October 2025. Is the lynx-ui repo truly requiring Node 24+ now? If so, that's fine, but it's a notable barrier for contributors since most people are on Node 20 or 22 LTS.

Minor: the `compare.png` asset is no longer referenced after this PR (removed from `introduction.mdx`, and `Elements&Components.mdx` is deleted). Could be cleaned up in a follow-up.

f0rdream:
> Good cleanup overall — the rewrite is much more focused and appropriately defers the full workflow to the upstream CONTRIBUTING.md. Deleting the WIP placeholders and fixing the `_meta.json` link from `/api/lynx-api/main-thread.html` to `/react/main-thread-script.html` are both correct.
> 
> Two items to check:
> 
> 1. Inconsistent script name — In both EN and ZH contributing docs, the "create" section uses `pnpm make-new-component --create` (kebab-case) while the "example" section uses `pnpm makeNewComponent --example` (camelCase). These should match whichever npm script name actually exists in the lynx-ui repo. If the script was renamed, the example command needs the same update.
> 2. Node.js >= 24.0.0 — The old docs specified >= 20. Node 24 was released April 2025 and won't reach LTS until October 2025. Is the lynx-ui repo truly requiring Node 24+ now? If so, that's fine, but it's a notable barrier for contributors since most people are on Node 20 or 22 LTS.
> 
> Minor: the `compare.png` asset is no longer referenced after this PR (removed from `introduction.mdx`, and `Elements&Components.mdx` is deleted). Could be cleaned up in a follow-up.

1. Fixed the inconsistency of pnpm scirpts.
2. The repo now requires Node 24, and Node 24 has been in LTS for near 1 year, so I documented like this. If we decide to change the repo requirement, I will update docs later, for now, I would keep 24

PR #969, opened 2026-05-07
docs: add Extracting Translations section to i18n docs

luhc228:
Change-Id: I82bda2aa682330c0962de7ff9185e1b4833fa2f8

- Add Extracting Translations section to i18n docs
- Document plugin-based extraction workflow using rsbuild-plugin-i18next-extractor (Node.js 22+), including lynx.config.ts setup, localesDir usage, importing JSON in src/i18n.ts, ignore options, and onKeyNotFound hook
- Describe CLI extraction workflow using i18next-cli with sample i18next.config.ts and extract command
- Introduce Lynx-specific deduplication flow using @lynx-js/i18next-translation-dedupe and runtime loadI18nextTranslations via customSections
- Provide the new guidance in both English and Chinese docs

### DIFF
--- a/docs/en/guide/inclusion/internationalization.mdx
+++ b/docs/en/guide/inclusion/internationalization.mdx
@@ -324,3 +324,176 @@ export function App() {
   img="https://lf-lynx.tiktok-cdns.com/obj/lynx-artifacts-oss-sg/lynx-website/assets/doc/i18n/switch-language.gif"
   entry="src/index.tsx"
 />
+
+## Extracting Translations
+
+If you downloaded many translations but your project only uses some of them, you may need to extract the translations to reduce the bundle size.
+
+There are two ways to extract the translations used in your source code.
+
+### Plugin (Recommended)
+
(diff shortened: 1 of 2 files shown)


Huxpro (changes requested):
Thanks for adding this section — the extraction workflow is a valuable addition to the i18n docs.

One issue that will break the setup for users:

### Missing `i18next-cli` peer dependency in install command

`rsbuild-plugin-i18next-extractor` declares `i18next-cli` as a **peer dependency** (not a regular dep):

```
peerDependencies: { "@rsbuild/core": "^1.0.0 || ^2.0.0-0", "i18next-cli": "^1.33.5" }
```

The upstream README correctly shows installing both:
```
npm add rsbuild-plugin-i18next-extractor i18next-cli --save-dev
```

But the PR's install command only installs the plugin:
```
<PackageManagerTabs command="install rsbuild-plugin-i18next-extractor -D" />
```

Users on pnpm/yarn (which don't auto-install peer deps) will hit errors. This should be:
```
<PackageManagerTabs command="install rsbuild-plugin-i18next-extractor i18next-cli -D" />
```

Applies to both the EN and ZH files.

---

Minor nit (non-blocking):

- The `onKeyNotFound` example shows `(key, locale) =>` but the actual callback signature has 4 parameters: `(key, locale, localeFilePath, entryName)`. The example works fine as-is (extra args are ignored in JS), but showing all four params would help users discover the full context available to them.

Huxpro (approved):

PR #974, opened 2026-05-07
feat(blur-view): add blur-view element with documentation and examples

MoonfaceX:
Change-Id: I408664747efcdc84088cbf74abfef94ba3c3faaa

### DIFF
--- a/docs/en/api/elements/built-in/_meta.json
+++ b/docs/en/api/elements/built-in/_meta.json
@@ -71,6 +71,11 @@
   },
   {
     "type": "file",
+    "name": "blur-view",
+    "label": "<blur-view>"
+  },
+  {
+    "type": "file",
     "name": "title-bar-view",
     "label": "<title-bar-view>"
   }
--- a/docs/en/api/elements/built-in/blur-view-API.mdx
+++ b/docs/en/api/elements/built-in/blur-view-API.mdx
@@ -1,0 +1,132 @@
+import { AndroidOnly, IOSOnly, ClayOnly, ClayWindowsOnly, ClayMacOSOnly, HarmonyOnly, Required, Deprecated, Go, PlatformTabs } from '@lynx';
+import { VersionBadge, APITable } from '@lynx';
+
+## Attributes
+
+### `android-capture-target`
+
+ <AndroidOnly /> <VersionBadge>3.9</VersionBadge>
+
+```tsx
+'android-capture-target'?: string;
+```
+
+The raw id of the Android Lynx view to capture and blur.
+Only the overlap with the 
+`<blur-view>`
+ is used as the blur source.
+When omitted or unresolved, Android uses the default parent capture path.
+
+
+### `blur-effect`
+
+ <IOSOnly />
+
+```tsx
+// @defaultValue: 'light'
+'blur-effect'?: 'light' | 'extra-light' | 'dark' | 'glass' | 'glass-container';
+```
+
+It mainly affects the brightness of the blurred area:
+light: the brightness is basically the same as the background;
+dark: the brightness is darker than the background;
+extra-light: the brightness is brighter than the background;
+glass: a visual effect that renders a glass material;
+glass-container: a UIGlassContainerEffect renders multiple glass elements into a combined effect.
+
+
+### `blur-radius`
+
+ <AndroidOnly /> <IOSOnly /> <HarmonyOnly />
+
+```tsx
+// @defaultValue: "0px"
+'blur-radius'?: string;
+```
+
+Gaussian blur radius.
+
+
+### `blur-sampling`
+
+ <AndroidOnly />
+
+```tsx
+// @defaultValue: 6
+'blur-sampling'?: number;
+```
+
+The downsampling ratio of the blurred area is similar to reducing the blur of the background area first and then zooming in,
+which slightly affects the blur effect, but can significantly improve performance.
+
+
+### `enable-auto-blur`
+
+ <AndroidOnly />
+
+```tsx
+// @defaultValue: true
+'enable-auto-blur'?: boolean;
+```
+
+Whether to automatically update blur.
+
+
+### `experimental-update-blur-radius`
+
+ <AndroidOnly /> <VersionBadge>3.4</VersionBadge>
+
+```tsx
+'experimental-update-blur-radius'?: boolean;
+```
+
+Switches the Android internal blur-buffer refresh path.
+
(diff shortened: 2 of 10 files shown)


Huxpro:
## Code Review

Thanks for adding `<blur-view>` docs! Below is a review of the diff.

### Overview
- New `blur-view.mdx` + `blur-view-API.mdx` in EN and ZH
- New compat-data file `packages/lynx-compat-data/elements/blur-view.json`
- Re-generated `packages/lynx-compat-data/api-stats.json` (4,334-line diff)
- Adds `@lynx-example/blur-view` to `packages/lynx-example-packages/package.json`
- Registers the page in EN/ZH `_meta.json` between `scroll-coordinator` and `title-bar-view`

### 🚩 Things to address

**1. Unrelated dependency change (likely accidental)**
In `packages/lynx-example-packages/package.json`:
```diff
-    "@lynx-example/viewpager": "0.6.12"
+    "@lynx-example/viewpager": "latest"
```
This is unrelated to blur-view, and every other entry in the file uses a pinned version. `"latest"` breaks build reproducibility. Please revert unless intentional — and if intentional, it deserves its own PR.

**2. EN/ZH structural drift**
The `:::info` callout and the basic-example heading appear in different orders:
- **EN** (`docs/en/.../blur-view.mdx`): `## Usage` → `### Basic` → `:::info` → `<Go ...>`
- **ZH** (`docs/zh/.../blur-view.mdx`): `## 使用指南` → `:::info` → `### 基础用法` → `<Go ...>`

Also EN says `## Usage` while ZH says `## 使用指南` (Usage Guide). Compare `title-bar-view.mdx` which uses `## Usage` ↔ `## 用法`. Pick one phrasing and one ordering and apply consistently.

**3. Same image used for two different examples**
Both **Basic** and **Liquid Glass For iOS** examples use the same `blur_view_base.jpg` (in EN and ZH). The Liquid Glass section should have its own preview asset.

**4. Unused MDX imports**
`blur-view.mdx` (EN and ZH) imports `AndroidOnly, Deprecated, IOSOnly, Required` but only uses `Go` and `APITable`. Trim them.

**5. Missing `<APISummary />`**
Sibling docs (`frame.mdx`, `title-bar-view.mdx`) render `<APISummary />` right after the H1. Adding it here would give readers a quick platform/version overview.

**6. Vague API descriptions**
- **`experimental-update-blur-radius`**: "Switches the Android internal blur-buffer refresh path." For an `experimental: true` flag this is too thin — when should a developer enable it? what does it trade off? Consider a `:::caution` callout.
- **`blur-radius`** is typed `string` with default `"0px"` but the doc doesn't specify accepted units (`px`, `rpx`?) or behavior on invalid input.
- **`glass-tint-color`** is typed `string` with default `'transparent'`; clarify accepted color formats (hex, rgba, named).
- **`spacing`** ("distance between elements at which they begin to merge") is opaque without noting it only applies when `blur-effect: 'glass-container'`. Same in ZH.
- **`android-capture-target`** mentions "raw id" without defining it — link to or briefly explain the Lynx Android raw-id concept.

**7. `flatten={false}` requirement is buried**
The fact that `flatten={false}` is required on the capture target is a real footgun. Consider promoting it next to the `android-capture-target` attribute description, not just the Usage info callout.

**8. `api-stats.json` regen scope**
The 4,334-line diff looks like a full regen. Worth confirming it was run from a clean tree so it doesn't sweep in unrelated changes from other elements.

**9. Empty PR description**
The PR body is just `Change-Id: …`. A short summary, screenshots, and a link to the example would help reviewers (and future archeologists).

**10. Failing Netlify preview**
The deploy preview is failing on `b5542a0`. Worth checking the build log before merge.

### 👍 What's good
- Compat-data file is thorough and platform-aware (Android / iOS / Harmony / Clay-Android / Clay-iOS rows filled; web / clay-windows / clay-macos correctly `false`).
- Version badges (`3.4`, `3.8`, `3.9`) on attributes match the compat-data versions — internally consistent.
- `_meta.json` registration is in sync between EN and ZH.
- The example package is alphabetized correctly.

### Recommendation
Request changes. Docs themselves are mostly in good shape, but the unrelated `viewpager` bump, the EN/ZH structural mismatch, the duplicated screenshot, and the failing Netlify build should all be resolved before merge. A pass over the experimental and platform-specific attribute descriptions for clarity would also help.

---
_Generated by [Claude Code](https://claude.ai/code)_

Huxpro (commented):
## Review: blur-view element docs

Good start on the blur-view documentation! A few items to address before this is merge-ready:

---

### Issues to fix

**1. `@lynx-example/blur-view` uses `"latest"` instead of a pinned version**
In `packages/lynx-example-packages/package.json`, every other dependency uses a pinned version (e.g. `"0.6.5"`). Using `"latest"` makes builds non-reproducible. Please pin to a specific version, or if the package isn't published yet, note it as a WIP blocker.

**2. Info block placement inconsistency between EN and ZH**
- EN (`blur-view.mdx`): the `:::info` block is _inside_ `### Basic`
- ZH (`blur-view.mdx`): the `:::info` block is _before_ `### 基础用法`, directly under `## 使用指南`

These should match structurally. I'd suggest placing it before the first subsection in both (matching the ZH version), since the info applies to all usage, not just the basic example.

**3. `android-capture-target` EN description has stray line breaks**
```
Only the overlap with the
`<blur-view>`
 is used as the blur source.
```
The `<blur-view>` reference is split across lines. Probably should be a single sentence: `Only the overlap with the \`<blur-view>\` is used as the blur source.`

**4. Both `<Go>` examples use the same preview image**
The Basic and "Liquid Glass For iOS" sections both link to `blur_view_base.jpg`. The Liquid Glass example should have its own screenshot showing the glass effect, otherwise users can't tell the difference.

---

### Suggestions (non-blocking)

**5. Consider adding `<APISummary />`**
Most element docs (view, scroll-view, image, overlay, text, title-bar-view, etc.) import and render `<APISummary />` after the heading. This gives readers a quick overview. Some XElement docs skip it (refresh, scroll-coordinator), so it's not strictly required, but it would be consistent with the majority.

**6. `blur-effect` value descriptions would benefit from a list format**
Currently the values are described as a single paragraph with semicolons. A markdown list would be more scannable:
```
- `light`: brightness is basically the same as the background
- `dark`: brightness is darker than the background
- ...
```

**7. Minor capitalization**: "Liquid Glass For iOS" — "For" is typically lowercase in titles ("Liquid Glass for iOS").

PR #991, opened 2026-05-11
refactor(lynx-ui): inline luna-packages as local components

Huxpro:
## Summary

- Move `packages/luna-packages/src/` → `src/luna/` as local components
- Promote `@dugyu/luna-*` dependencies to root `package.json`
- Remove `@lynx-js/luna-packages` workspace package entirely
- Update all path aliases (`@luna` in rspress.config.ts, tsconfig.json)
- Update tailwind content scan path and luna-demo script defaults

## Motivation

`@lynx-js/luna-packages` was the only `workspace:*` dependency in `dependencies` (not `devDependencies`). This breaks downstream consumers that install `lynx-website` as a git dependency — pnpm cannot resolve `workspace:*` references outside the source workspace.

The package was only 4 source files exporting a single `LunaStudio` component, used in one MDX page (`luna-themes-tokens.mdx`). It doesn't need to be a separate workspace package — local components under `src/` are sufficient.

## Test plan

- [ ] `pnpm install` succeeds without workspace resolution errors
- [ ] `pnpm build` completes successfully
- [ ] Luna themes tokens page renders correctly with LunaStudio component
- [ ] TypeScript compilation passes (`tsc --noEmit`)

Inline Luna packages as local components by migrating workspace package to src/luna

**Changes:**
- Move Luna source files from `packages/luna-packages/src/` to `src/luna/` as local components (index.ts, interaction.ts, layout.ts, luna-studio.tsx)
- Add `@dugyu/luna-demo-bundles` (0.1.0), `@dugyu/luna-stage` (0.1.1), and `@dugyu/luna-studio` (0.1.1) to root `package.json` dependencies
- Remove `@lynx-js/luna-packages` workspace package entirely, including its README and package.json manifest
- Update `@luna` path alias in `rspress.config.ts` to resolve to `src/luna/index.ts` instead of `packages/luna-packages/src/index.ts`
- Update `@luna` path mappings in `tsconfig.json` to point to `./src/luna/...` instead of `./packages/luna-packages/src/...`
- Update Tailwind `content` configuration in `tailwind.config.js` to scan `./src/luna/**/*.{js,ts,jsx,tsx}` instead of `./packages/luna-packages/src/**/*.{js,ts,jsx,tsx}`
- Change default `LUNA_SOURCE_DIR` in `scripts/luna-demo.js` from `packages/luna-packages/node_modules` to `node_modules`

**Motivation:**
[...]

PR #992, opened 2026-05-11
fix: replace catalog: protocol with explicit versions

Huxpro:
## Summary

- Replace all `catalog:` protocol references in root `package.json` with explicit version specifiers from `pnpm-workspace.yaml`
- Same class of issue as `workspace:*` fixed in #991 — `catalog:` is not resolved when the package is consumed as a git dependency

## Changed dependencies

| Package | Before | After |
|---------|--------|-------|
| `@lynx-js/web-core` | `catalog:` | `^0.20.3` |
| `motion` | `catalog:` | `^12.23.0` |
| `react` | `catalog:` | `^18.3.0` |
| `react-dom` | `catalog:` | `^18.3.0` |
| `@types/react` | `catalog:` | `^18.3.0` |
| `@types/react-dom` | `catalog:` | `^18.3.0` |

The catalog entries in `pnpm-workspace.yaml` are preserved — workspace packages can still use `catalog:`.

## Test plan

- [x] `pnpm install` succeeds
- [ ] CI passes


Huxpro:
Follow-up: #993 removes the stale `catalog:` block from `pnpm-workspace.yaml` — now that all `catalog:` protocol references are replaced with explicit versions, the catalog definition itself is unused.

PR #999, opened 2026-05-12
fix: go example github link

f0rdream:
Change-Id: I3de96d306759c84fbea83de1591cbe241244bbfe

Update `prepare:lynx-ui-example-data` script to reference lynx-ui example packages from node_modules

- Update `EXAMPLES_DIR` path in `prepare:lynx-ui-example-data` npm script from `apps/examples` to `packages/lynx-ui-example-packages/node_modules/@lynx-example`

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-website/pull/999)


Huxpro (approved):
LGTM. The fix is correct.

The `repository.directory` field in each lynx-ui example package already contains the full relative path (e.g. `apps/examples/Button`). Having `apps/examples` also in `EXAMPLE_GIT_BASE_URL` caused a doubled path segment, producing broken GitHub links like `.../tree/main/apps/examples/apps/examples/Button`. Removing the suffix from the base URL fixes this.

PR #1019, opened 2026-05-18
fix(lynx-ui-home): reduce LunaStudioShowcase theme flash by relying on CSS tokens

Dugyu:
- Avoid initial light→dark flash on direct `/lynx-ui` entry by styling `LunaStudioShowcase` (locked mode) with Rspress CSS variables instead of `useDark()`-driven hardcoded colors.
- Introduce `canvas` (`bg-canvas-ambient`) and `rule` (`bg-rule`) utilities mapped to `--rp-c-bg-alt` / `--rp-c-divider-secondary` for stable, theme-aware surfaces and dividers.
- Define lynx-ui subsite token overrides for `--rp-c-bg-alt` and `--rp-c-divider-secondary` and scope `--tw-ring-color` to prevent Tailwind preflight from overriding focus ring color.

Change-Id: I508de2070cd1af9824ef36030fc72bdc660048f8

## Reduce theme flash by relying on CSS tokens in lynx-ui

- Update `LunaStudioShowcase` to style locked mode with Rspress CSS variables instead of useDark()-driven hardcoded colors, preventing initial light→dark flash on `/lynx-ui`
- Add `canvas` and `rule` Tailwind utility classes mapped to `--rp-c-bg-alt` and `--rp-c-divider-secondary` CSS variables for consistent, theme-aware surfaces and dividers
- Centralize divider styling in controls panel to use `bg-rule` utility instead of inline conditional theme colors
- Update `IconToggleButton` focus ring styling to use `focus-visible:ring-offset-background` token for both light and dark cases
- Provide `lynx-ui` subsite CSS variable overrides for `--rp-c-bg-alt`, `--rp-c-text-1`, and `--rp-c-divider-secondary` in both light and dark modes
- Scope `--tw-ring-color` in `lynx-ui` theme to prevent Tailwind preflight from overriding focus ring color styling

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-website/pull/1019?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/luna/luna-studio.tsx
+++ b/src/luna/luna-studio.tsx
@@ -125,8 +125,8 @@ function IconToggleButton(props: {
       className={cn(
         'h-9 w-9 rounded-full border transition-all hover:scale-110 focus-visible:ring-2 focus-visible:ring-offset-2',
         isLight
-          ? 'border-black/10 focus-visible:ring-black/30 focus-visible:ring-offset-white'
-          : 'border-white/10 focus-visible:ring-white/40 focus-visible:ring-offset-black',
+          ? 'border-black/10 focus-visible:ring-black/30 focus-visible:ring-offset-background'
+          : 'border-white/10 focus-visible:ring-white/40 focus-visible:ring-offset-background',
         props.active ? activeClassName : inactiveClassName,
         props.className,
       )}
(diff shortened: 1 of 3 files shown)


Huxpro (approved):
APPROVE — the approach is correct: replacing JS-driven hardcoded colors with CSS custom properties for locked/page-integrated mode should eliminate the hydration-dependent theme flash, and the token values match the old hardcoded values closely enough to avoid visual regression.

Minor non-blocking notes:

1. `theme/index.scss`: the standalone `--tw-ring-color` declarations are immediately overridden by the following nested `&, & *, & *::before, & *::after` rule, so they appear to be dead code. Consider removing them to avoid confusion.

2. `src/luna/luna-studio.tsx`: `transition-all` is broader than needed for a theme/color fix and may animate layout-related properties in future changes. `transition-colors duration-300` would be safer.

3. The divider change to `bg-rule` is unconditional while container/control backgrounds branch on `themeModeLocked`. This is fine because the token values match the old `bg-black/10` / `bg-white/10`, but a small comment could help future readers.

What looks good:
- lynx-ui token overrides are correctly scoped for light/dark.
- `bg-soft` is already an established project utility and fits the controls background.
- `focus-visible:ring-offset-background` simplifies duplicated conditional classes.
- `canvas` / `rule` Tailwind mappings are clean semantic token aliases.

Huxpro (approved):
APPROVE — the approach is correct: replacing `useDark()`-driven hardcoded colors with CSS custom properties for the locked/page-integrated mode should eliminate the hydration-dependent light→dark flash. The SCSS token values match the old hardcoded values, so I don't see a visual regression in the fallback path.

Non-blocking notes:

1. `theme/index.scss`: the standalone `--tw-ring-color` declarations look dead because they are immediately overridden by the nested `&, & *, & *::before, & *::after` rule in both light and dark blocks. Consider removing the standalone declarations to avoid confusion.
2. `src/luna/luna-studio.tsx`: `transition-all` is broader than needed for a theme-flash fix. `transition-colors duration-300` would avoid accidentally animating future layout/size/shadow changes.
3. `dividerClassName = 'bg-rule'` is unconditional while the other classes branch on `themeModeLocked`. This is fine because the token values match the previous `bg-black/10` / `bg-white/10`, but a short comment could make the intent clearer.

What looks good:
- `--rp-c-bg-alt`, `--rp-c-text-1`, and `--rp-c-divider-secondary` are correctly scoped to `[data-subsite='lynx-ui']` in both light and dark.
- `bg-soft` is already an established project utility, so using it for the controls background aligns with the theme system.
- `focus-visible:ring-offset-background` removes a conditional that was effectively doing the same thing in both branches.
- The Tailwind `canvas` / `rule` semantic mappings are clean.

PR #1032, opened 2026-05-21
docs: refine lynx-ui motion story

f0rdream:
Refines the lynx-ui blog flow around low-latency interaction, MTS, and Motion.

Adds Swiper and Sheet demos to connect the technical architecture with the component storytelling, and moves the broader component catalog into an archived section so the launch narrative stays focused.

### DIFF
--- a/docs/en/blog/lynx-ui.mdx
+++ b/docs/en/blog/lynx-ui.mdx
@@ -105,45 +105,11 @@ But products are not built from atoms alone. That is where the [frontend compone
   webPreviewMode="auto"
 />
 
-## Headless Components
+## Composability and Unlocking Design Systems
 
 Abstraction only becomes truly useful when it stays composable. lynx-ui does not treat components as black boxes with endlessly growing prop surfaces. Instead, it exposes structured primitives that let teams decide how pieces fit together in their own product context. A dialog is not a single opaque component, but a composition of roles such as trigger, content, and state boundary. This model unlocks more than flexibility. It gives teams a stable way to compose behavior, structure, and state without turning components into rigid black boxes. lynx-ui provides the interaction logic and native integration, while leaving product teams free to decide how each piece should be assembled in real scenarios.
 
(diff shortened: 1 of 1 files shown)


Huxpro (commented):
## Review: docs: refine lynx-ui motion story

**Scope**: Commit 3 (`f46e5f7`) — the actual delta on top of the base branch. Commits 1–2 are already merged into `feat/lynx-ui-blog`.

### What this does well

- Resolves two earlier TODOs by replacing the static Popover code block and the Swiper placeholder with real `<Go>` interactive examples. The blog post now *shows* what it describes.
- The narrative flow from composability → low-latency interaction → motion → CTA reads much better. Making Motion a subsection of Low-Latency Interaction is structurally sound — it positions motion as a consequence of the architecture, not a separate feature.
- Adding the Sheet example as a companion to Swiper strengthens the story: Swiper shows continuous tracking, Sheet shows spring-based settling. Together they demonstrate both sides of the MTS + Motion pairing.

### Issues to address

**1. Lost hyperlinks for dual-thread model and MTS** _(medium)_

The base version linked both concepts:
```md
[dual-thread model](/guide/scripting-runtime/index.html#javascript-runtime)
[MTS](/react/main-thread-script.html)
```

The new text mentions both inline but without links:

> Lynx's dual-thread model gives lynx-ui a way to handle that tradeoff. [...] MTS can move the high-frequency part of the interaction onto the main thread

For a blog post introducing lynx-ui to new readers, these are the two most important architectural concepts in the section — readers need a path to learn more. Please restore the links.

**2. TODO left in source** _(low)_

```
{/* TODO: Insert a `@lynx-js/motion` final doc link */}
```

Invisible to readers (MDX comment), so not a launch blocker. But if the doc link is expected before the blog goes live, this should be tracked somewhere more visible than inline source.

**3. Sponsorship mention removed** _(informational)_

The old text included "We're now sponsoring the Motion.js project" and "Same code. Same API. Web and native." These are strong trust-building signals for an OSS audience. Intentional removal is fine if the intent is to streamline, but worth a conscious decision rather than an accidental drop.

### Verdict

**LGTM with minor fixes.** The structural improvements and interactive examples are a clear upgrade. Restoring the dual-thread/MTS links is the one thing I'd want to see before this merges — the rest is discretionary.

PR #1047, opened 2026-05-26
docs: lynx-fiber-element skill

gaoachao:
## Summary
- add English and Chinese docs pages for the new `lynx-fiber-element` skill
- document when to use the skill, its bundled references/examples, and the recommended workflow
- add the new skill page to the AI docs navigation in both locales

## Why
`lynx-fiber-element` is a new skill in `lynx-community/skills`, and the Lynx docs should expose it alongside the existing AI skills.

## Validation
- `pnpm run build`

- Add documentation pages for `lynx-fiber-element` skill in English and Chinese
- Register `lynx-fiber-element` in navigation metadata for both English and Chinese docs

[Review Change Stack](https://app.coderabbit.ai/change-stack/lynx-family/lynx-website/pull/1047?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/en/ai/_meta.json
+++ b/docs/en/ai/_meta.json
@@ -36,6 +36,10 @@
   },
   {
     "type": "file",
+    "name": "skills/lynx-fiber-element.mdx"
+  },
+  {
+    "type": "file",
     "name": "skills/lynx-typescript.mdx"
   },
   {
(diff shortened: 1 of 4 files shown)


Huxpro (changes requested):
We should not use the word "fiber" within any public docs or as public API

PR #1058, opened 2026-05-29
docs: @lynx-js/genui api

gaoachao:


Huxpro:
## Investigation summary (CI failures)

Looking into why Netlify deploy-preview and the three Cloudflare Pages projects (`lynx-website-next`, `lynx-website-release-3-4`, `lynx-website-release-3-5`) all fail while GitHub Actions "Validate" passes.

### What we know

- "Validate" runs `pnpm run build` with `NODE_OPTIONS=--max-old-space-size=4096 RSPRESS_SSG_WORKER_THREAD_COUNT=2` and passes (~5 min).
- Netlify deploy-preview runs `--max-old-space-size=3072 RSPRESS_SSG_WORKER_THREAD_COUNT=1` with `RSPRESS_LIGHTWEIGHT_BUILD=true` (per #1059). Fails at ~3.5 min.
- The three CF Pages projects don't read `netlify.toml`, so they run the default `pnpm run build` (same command as Validate). They fail at 9 / 17 / 26 min.
- A snapshot of one failing build shows the `build` step stuck on a spinner for 22h+ — i.e. **hung, not crashed**. This rules out heap OOM (which would crash within minutes).
- This PR adds 626 new `.md` files under `docs/en/api/genui/` (+21k lines, ~35% increase over the existing `docs/en/api/*` baseline of ~1800 pages).

### Working hypotheses, ranked

1. **SSG worker deadlock or serialization stall** with `RSPRESS_SSG_WORKER_THREAD_COUNT=1` on the larger page set. The single-worker setting was meant to ease memory pressure (#1059) but may be exposing a serialization issue at this volume.
2. **`pluginLLMsPostprocess` or dead-link checking** stalling. These run on the non-lightweight CF builds and traverse all pages. With 2,400+ pages and many cross-references, a quadratic step could blow past the build timeout.
3. **A specific page** triggering an infinite loop in a markdown / remark / shiki transformer.

### Why this isn't ready to merge

- All three Cloudflare Pages projects auto-deploy `main`. They're failing on the PR head using the **same build command** they'll run after merge, so merging this is very likely to break the live site rather than fix it.
- The four red checks are consistent across two independent providers (Netlify + Cloudflare). That's hard to dismiss as a preview-only quirk.
- We have not confirmed the root cause. Merging now would be a bet on the failure being unrelated to this PR's content — without evidence to support that bet.

### What we need to narrow it down

The last log line before the hang in any one of the four failing builds. That tells us which phase / plugin / page is the culprit. If anyone with dashboard access can grab the tail of the Netlify deploy-preview log (or run `wrangler pages deployment tail` for one CF project), the fix should be targeted.

Until then, candidate mitigations are all guesses:
- Restore `RSPRESS_SSG_WORKER_THREAD_COUNT=2` in `netlify.toml` (if it's a 1-worker-only issue).
- Add `**/api/genui/**` to `route.exclude` when `IS_LIGHTWEIGHT_BUILD` (cuts the new content from previews; does **not** help CF since they don't run lightweight).
- Temporarily disable `pluginLLMsPostprocess` to test.

Huxpro:
Merged — thank you @gaoachao for landing this massive surface, 626 en + 626 zh genui API pages! 🎉

For the record on the failing checks:

- `Validate` / `Done` started failing after #1086 landed an unformatted `scripts/ci/check-build-prereqs.mjs` on main. Not introduced by this PR; will follow up with a `prettier --write` fix on main so future PRs aren't blocked by it.
- Netlify deploy-preview and the three Cloudflare Pages projects fail with the lightweight build at this content volume. We reproduced the build locally (with `RSPRESS_LIGHTWEIGHT_BUILD=true`) and it completes successfully at ~7 GB peak RSS, which overshoots Netlify's 3 GB heap cap. The previously-committed sidebar tweaks above are unrelated to this; the OOM ceiling is the next thing to chase.

Merging via admin override since both failure modes are pre-existing/infrastructure and the content itself is good.

PR #1070, opened 2026-06-02
fix release/3.8

mitchilling:
- next/blog link
- first page route
- 3.7

Change-Id: Ibc0269259bbade89077f5b6d590790d0fc409418


Dugyu (changes requested):
We also need to set  base: `/${versionJson.current_version}` to base: `/` in `rslib.config.ts`

Dugyu (approved):

PR #1075, opened 2026-06-02
fix(border-beam): inherit host border-radius to stop beam leaking past corners

Huxpro:
## Summary

- `BorderBeam` hardcoded `border-radius: 24px` on its frame, while the active card hosts (`.choice-tabs__card--active`, `.shared-tabs__card--active`, used by ChoiceTabs / PlatformTabs / TutorialTabs) use `14px`. The frame's `overflow: hidden` therefore clipped at a different curve than the visible card, and the rectangular canvas stroke bled past the rounded corners.
- Symptom: a stray red/teal gradient streak orbiting just outside the active ChoiceTab card (most visible on `/next/guide/start/quick-start.html`), and the same on PlatformTabs.
- Fix: change the frame to `border-radius: inherit` so the clip always matches whatever host renders the beam.

## Before / after

The beam orbits the perimeter every 3s. In the buggy state, the beam visibly poked outside the active card at each corner. After the fix, the beam stays inside the rounded card on both ChoiceTabs and PlatformTabs (verified via 3 sequential frame captures per page).

Computed `border-radius` on `.border-beam-frame` after the fix:
- `.choice-tabs__card--active` host → 14px (was 24px)
- `.shared-tabs__card--active` host → 14px (was 24px)

## Test plan

- [ ] Hit `/next/guide/start/quick-start.html`, look at the "Try with Explorer" active card across one full 3s animation cycle: no colored streak outside the card
- [ ] Hit `/next/guide/use-data-from-host-platform.html`, same check on the active PlatformTab
- [ ] Spot-check homepage feature cards (other BorderBeam consumers) — they already used 14-ish radii via design tokens, so no regression expected

Inherit host border-radius in BorderBeam to prevent beam leakage

- Set border-radius: inherit on .border-beam-frame so the canvas clip matches the host card corners and prevents the beam from bleeding past rounded edges.
- Scale canvas by devicePixelRatio and use CSS-pixel dimensions to avoid HiDPI subpixel artifacts when drawing.
- Read the corner radius from the frame’s computed style so the beam path follows the host’s actual radius.
- Trace an inset rounded-rectangle path (four edges + four quarter-circle arcs) and inset by lineWidth/2 so the stroke sits fully inside the clipped area.
- Map perimeter distance to coordinates with a segment/arc approach and draw the beam by sampling along that path to produce a continuous stroke around the inset rounded rect.
- Replace previous gradient with a defined sequence of color stops to produce a teal head and red body that fade to transparent at both ends.
- Remove remaining 1–2px streaks where straight strokes met clipped corners by aligning stroke geometry with the host clipping and HiDPI scaling.

### DIFF
--- a/src/components/home-comps/border-beam/index.module.less
+++ b/src/components/home-comps/border-beam/index.module.less
@@ -6,7 +6,11 @@
   width: 100%;
   height: 100%;
   overflow: hidden;
-  border-radius: 24px;
+  // Match the host card's rounded corners so the rectangular canvas
+  // stroke is clipped at the same curve. A hardcoded radius would let
+  // the beam bleed past the parent's corners on any host that doesn't
+  // use the same value.
+  border-radius: inherit;
   box-sizing: border-box;
   z-index: -1;
   pointer-events: none;
--- a/src/components/home-comps/border-beam/index.tsx
+++ b/src/components/home-comps/border-beam/index.tsx
@@ -25,6 +25,15 @@ const BorderBeam: React.FC<BorderBeamProps> = ({
     const ctx = canvas.getContext('2d', { alpha: true });
     if (!ctx) return;
 
+    // CSS pixels — kept separate from canvas.width/height so HiDPI scaling
+    // doesn't distort the drawn path.
+    let cssWidth = 0;
+    let cssHeight = 0;
+    // Corner radius inherited from the host card. Keeping the beam path on
+    // the same curve the host clips to is what stops it bleeding past the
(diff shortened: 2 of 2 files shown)


Huxpro:
Closing — the BorderBeam corner-leak fix turned out to be more invasive than the visual was worth (the beam is barely perceptible on these tab cards anyway). The real, observable bug in this area is the active card's drop shadow being clipped by the carousel track on narrow widths. Tracking that separately.

PR #1090, opened 2026-06-04
feat(badge): use Rspress icons for Experimental and Deprecated, dedup on react-compiler

Huxpro:
## Summary

Two related changes to make the experimental/deprecated badges consistent across the site.

### 1. `<Experimental />` and `<Deprecated />` now use Rspress's `Tag`

`src/components/api-badge/StatusBadge.tsx` previously rendered raw `<Badge text="..." type="...">` chips with no icon. Rspress's built-in `<Tag>` already renders "experimental" and "deprecated" with the flask / deprecated icon and is what fires when you set `tag: experimental` in frontmatter. Delegating our wrappers to `<Tag>` means every inline `<Experimental />` and `<Deprecated />` picks up the same icons for free, and the chip looks identical whether the marker came from frontmatter or from MDX.

Side effect: the label lowercases ("experimental", "deprecated") to match Rspress's convention. The unused `since` prop on `<Deprecated />` is dropped (zero call sites in docs).

### 2. Drop the duplicated badge on react-compiler

`docs/{en,zh}/react/react-compiler.mdx` had both `tag: experimental` frontmatter AND an inline `<Experimental />` on the H1, so the heading was rendering two chips side-by-side. Removed the inline component and its import — the frontmatter tag alone covers heading + sidebar.

## What this does NOT change

`<Experimental />` stays exported and is still used inline by:
- `docs/{en,zh}/blog/lynx-3-5.mdx` — section-level annotation on a release-notes section
- `docs/{en,zh}/help/components.mdx`, `docs/zh/internal/docs-contribution-guide/doc-components.mdx` — demo / reference uses

Those are not page-level experimental markers, so frontmatter would be wrong for them. They still work and now render with the Rspress icon.

## Test plan

- [ ] Netlify deploy preview: `/react/react-compiler` (en + zh) shows one experimental chip in the H1 area and one in the sidebar nav
- [ ] Every inline `<Experimental />` and `<Deprecated />` on other pages renders with the Rspress flask / deprecated icon
- [ ] Sidebar still shows the flask next to "React Compiler"

### DIFF
--- a/docs/en/react/react-compiler.mdx
+++ b/docs/en/react/react-compiler.mdx
@@ -4,9 +4,8 @@ description: 'Use React Compiler to auto-optimize components in ReactLynx.'
 ---
 
 import { PackageManagerTabs } from '@theme';
-import { Experimental } from '@lynx';
 
-# React Compiler <Experimental />
+# React Compiler
 
 [React Compiler](https://react.dev/learn/react-compiler) can optimize components at compile time, eliminating the need to manually use `useMemo`, `useCallback`, and `memo`. You can use React Compiler in ReactLynx to enhance your application's performance.
 
--- a/docs/zh/react/react-compiler.mdx
+++ b/docs/zh/react/react-compiler.mdx
@@ -4,9 +4,8 @@ description: '用 React Compiler 自动优化组件。'
 ---
 
 import { PackageManagerTabs } from '@theme';
-import { Experimental } from '@lynx';
 
-# React Compiler <Experimental />
+# React Compiler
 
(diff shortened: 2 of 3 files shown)


SoonIter:
Thanks for the fix! This makes the badge rendering more consistent.

One related follow-up: after the badge is rendered in the H1, the Algolia crawler may incorrectly include the badge text in the extracted title, so the search result can become something like `React Compiler Experimental experimental`.

image


We should update the Algolia crawler config to remove badge/non-doc elements before extracting headings, similar to the Rspress example:
https://rspress.rs/plugin/official-plugins/algolia#algolia-crawler-config

```js
// Remove badge elements to prevent their text from being indexed
$('.rp-badge').remove();
// Remove non-doc elements in h1 to keep title text clean
$('.rspress-doc h1 .rp-not-doc').remove();
```

PR #1117, opened 2026-06-15
fix(ci): unblock API docs auto-PR on cspell pre-commit failure

upupming:
## Background

The weekly **Update API docs** workflow ([run that failed](https://github.com/lynx-family/lynx-website/actions/runs/27507979274/job/81302597040)) failed at the **Create Pull Request** step.

`peter-evans/create-pull-request` runs its own `git commit`, which triggers the husky `pre-commit` hook (`lint-staged` → `cspell`/`prettier`). On the freshly generated genui docs, `cspell` flagged the identifier `openui` as an unknown word:

\`\`\`
docs/en/api/genui/genui.openuirendererruntimeprops.md:164:5 - Unknown word (openui)
docs/en/api/genui/genui.openuirendererruntimeprops.response.md:7:5 - Unknown word (openui)
docs/zh/api/genui/...  (same 2)
husky - pre-commit script failed (code 1)
\`\`\`

The hook failed → the action aborted → no PR was opened.

## Changes

1. **`cspell.json`** — add `openui` to the dictionary (next to the existing `genui`); it's a legitimate auto-generated API identifier.
2. **`.github/workflows/update-api-docs.yml`** — set `HUSKY=0` on the *Create Pull Request* step so the action's internal commit skips the pre-commit hook. This is the documented husky equivalent of `git commit -n`. The auto-generated docs are already linted/prettified by the regeneration script, so running the staged-files hook again on this bot commit adds no value and only risks blocking the run on harmless new identifiers in the future.

Disable Husky pre-commit hooks in the Create Pull Request step of the update-api-docs workflow to unblock CI on auto-generated API docs  
Add a hard-gated “Validate generated docs” step that runs `scripts/ci/check-asset-urls.mjs` for `docs/en/api` and `docs/zh/api` before creating the PR  
Add `openui` to the `cspell.json` dictionary to prevent spell-check failures on generated API identifiers

### DIFF
--- a/.github/workflows/update-api-docs.yml
+++ b/.github/workflows/update-api-docs.yml
@@ -74,6 +74,12 @@ jobs:
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7
+        # The action runs its own `git commit`, which triggers the husky
+        # pre-commit hook (lint-staged → cspell/prettier). On auto-generated
+        # docs that hook can fail on harmless unknown words and abort the run.
+        # HUSKY=0 skips the hook for this commit — equivalent to `git commit -n`.
+        env:
+          HUSKY: '0'
         with:
           branch: bot/update-api-docs
           delete-branch: true
--- a/.github/workflows/update-api-docs.yml
+++ b/.github/workflows/update-api-docs.yml
@@ -72,6 +72,14 @@ jobs:
         continue-on-error: true
         run: pnpm run build
 
+      - name: Validate generated docs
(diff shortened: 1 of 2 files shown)


upupming:
Good catch — fixed in 68eca425. `HUSKY=0` does skip the whole hook, so I added an explicit **Validate generated docs** step that runs `node scripts/ci/check-asset-urls.mjs` against `docs/{en,zh}/api` before the PR is created. It's a hard gate (no `continue-on-error`), so blocked internal asset URLs will fail the run rather than reach the draft PR. cspell/prettier stay skipped since they're just noise on auto-generated docs.

PR #1159, opened 2026-07-07
perf(rspress): skip generated HTML sanitizing in dev for build perf

SoonIter:
## Summary
- Skip generated HTML sanitizing work when Rspress is not running a production build, avoiding the post-build HTML scan during dev.
- Type the local Rspress plugin and `fs.readdir` entries explicitly so the config remains TypeScript-checkable.

## Testing
- pnpm exec prettier --check rspress.config.ts
- pnpm exec tsc --noEmit --pretty false --skipLibCheck --moduleResolution bundler --module ESNext --target ES2020 --lib DOM,ES2021 --jsx react-jsx --resolveJsonModule --allowImportingTsExtensions --strict --types node rspress.config.ts

- Skip HTML sanitizing during non-production builds
- Type the local Rspress plugin with `RspressPlugin`
- Tighten `fs.readdir` entry typing with `Dirent`


SoonIter:
LGTM, small affect so quick merge

PR #1220, opened 2026-07-27
fix(version): add urgent root redirect workaround for hydration

SoonIter:
## Summary

This is an urgent workaround for the production homepage hydration failure. The 4.0 release is built with the `/4.0/` Rspress base, but Netlify serves the same output at `/`, so the client router cannot match its basename. Temporarily redirect `/` to `/4.0/` with a 302 to align the browser pathname and restore hydration. A follow-up can revisit serving the latest release at `/` by building it with a root base.

## Related Links

- https://github.com/lynx-family/lynx-website/pull/1212

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -185,10 +185,10 @@
   to = "https://release-3-4--lynx-doc.netlify.app/:splat"
   status = 200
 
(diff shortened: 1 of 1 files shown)


Huxpro (approved):
Thank you!

PR #1231, opened 2026-07-28
fix(version): redirect bare version roots with 308 `/next` -> `/next/`

SoonIter:
## Summary

- Redirect bare `/4.0` and `/next` version roots to their trailing-slash forms with 308 responses from a Netlify Edge Function, avoiding redirect-rule URL normalization loops.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,6 +5,19 @@
 [context.deploy-preview.environment]
   RSPRESS_LIGHTWEIGHT_BUILD = "true"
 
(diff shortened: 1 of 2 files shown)


SoonIter:
Related: #1229 — this PR is the server-side counterpart: trailing-slash canonicalization for bare version paths (`/next`, `/4.0`, `/3.4`–`/3.9`) is enforced at the edge via explicit 308 rules, instead of relying on client-side handling.

Ref: [Why does the trailing slash matter? A short history](https://kentgigger.com/posts/why-does-the-trailing-slash-matter-a-short-history)

PR #1233, opened 2026-07-30
docs: clarify event-through and pan interception

rAY-ooo:
## Summary

- Clarify `event-through` inheritance, `<page>` behavior, hit-testing dependencies, and client integration on Android, iOS, and HarmonyOS.
- Document the iOS-only filtered pan-interception attributes and mark the shared pan-interception attributes as Android/iOS only.
- Add iOS 4.1 compatibility data for the filtered pan-interception attributes and regenerate API statistics.

## Why

The existing documentation did not fully describe the `<page>` inheritance boundary, the interaction with `enableEventThrough`, or how the hit-tested node affects `event-through`. It also omitted the iOS filtered pan-interception attributes.

## Upstream source

- [lynx-family/lynx#8483](https://github.com/lynx-family/lynx/pull/8483) adds the durable `typePackageSince` metadata to the config source and regenerates the `@lynx-js/type-config` TSDoc declaration.
- The EN/ZH canonical `enableEventThrough` pages in this PR were regenerated with API Extractor/API Documenter from that upstream TSDoc; the generated output matches both files byte-for-byte.

## Impact

Developers can determine event pass-through behavior more accurately and identify the platform/version requirements for the pan-interception APIs.

## Validation

- Prettier
- CSpell
- API documentation checker
- Compatibility-data schema and key-name validation
- Compatibility-data unit tests (10 tests)
- Pre-commit asset URL and API summary checks
- Upstream config generator and Python unit tests
- API Extractor/API Documenter regeneration from upstream TSDoc (EN/ZH output matches byte-for-byte)

- Clarify `event-through` inheritance, `<page>` behavior, hit-testing dependencies, and platform integration
- Document Android/iOS pan-interception support and iOS-specific filtering attributes
- Record iOS 4.1 compatibility for pan-interception attributes
- Regenerate API statistics and validate documentation and compatibility data

### DIFF
--- a/docs/en/api/elements/built-in/view.mdx
+++ b/docs/en/api/elements/built-in/view.mdx
@@ -515,7 +515,7 @@ Specify whether Lynx consumes touch events from the platform layer when a touch 
 
 For non-`<page>` nodes, this property supports inheritance. If the hit-tested node does not set `event-through`, it inherits the `event-through` value from its ancestors, excluding the `event-through` value of the `<page>` node.
 
-For the `<page>` node, event pass-through is determined jointly by its `event-through` property and the page configuration [`Config.enableEventThrough`](https://lynxjs.org/api/rspeedy/type-config.config.enableeventthrough). When a touch hits the `<page>` node's area, Lynx does not consume the platform-level touch event if either value is `true`. This configuration property is available in `@lynx-js/type-config` 4.1.0 or later; this package floor is independent of its LynxSDK 3.2 runtime floor.
+For the `<page>` node, event pass-through is determined jointly by its `event-through` property and the page configuration [`Config.enableEventThrough`](../../rspeedy/type-config.config.enableeventthrough.md). When a touch hits the `<page>` node's area, Lynx does not consume the platform-level touch event if either value is `true`. This configuration property is available in `@lynx-js/type-config` 4.1.0 or later; this package floor is independent of its LynxSDK 3.2 runtime floor.
 
 The `event-through` result is based on the node hit by the touch. [`user-interaction-enabled`](#user-interaction-enabled) and [`pointer-events`](../../css/properties/pointer-events.mdx) affect hit testing and the resulting hit node, so they can indirectly affect whether `event-through` takes effect. After the hit node is determined, Lynx applies the `<page>` rules above if the hit node is `<page>`; otherwise, it evaluates the `event-through` values of the hit node and its non-`<page>` ancestors.
 
@@ -525,7 +525,16 @@ The active region of this property is affected by [`event-through-active-regions
 `event-through` only controls whether Lynx responds to or consumes touch events from the platform layer. Passing unconsumed touch events to other platform-layer nodes also requires client-side integration for each platform:
 
 - **Android**: Ensure that the parent container of the Lynx page does not consume touch events. When Lynx leaves an event unconsumed, the system's event dispatch can pass it to a sibling node that consumes it, such as a scroll container.
-- **iOS**: When LynxView does not respond to a touch, its `hitTest` returns `nil` and the Lynx page is excluded from the event responder chain. To let a sibling node respond to the event, the parent container's `hitTest` must also return `nil` as appropriate.
+- **iOS**: When LynxView does not respond to a touch, its `hitTest` returns `nil` and the Lynx page is excluded from the event responder chain. The following example assumes that a dedicated Lynx wrapper contains LynxView and overrides `hitTest:withEvent:`. The intended responder is a sibling of that wrapper under their common ancestor:
+
+  ```text
+  Common ancestor
+  ├── Lynx wrapper (overrides hitTest:withEvent:)
+  │   └── LynxView
+  └── Intended responder view
+  ```
+
+  When LynxView does not respond, the wrapper also returns `nil`, allowing the common ancestor to continue hit testing the wrapper's sibling. If the intended responder is instead a sibling of LynxView inside the wrapper, do not use this override unchanged: it replaces a non-Lynx descendant returned by `[super hitTest:withEvent:]` with `nil`.
 
   ```objc
   - (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event {
--- a/docs/en/api/elements/built-in/view.mdx
+++ b/docs/en/api/elements/built-in/view.mdx
@@ -378,7 +378,7 @@ On the Android platform, this setting only applies to `view` elements, and the c
 Gesture interception effects only apply to sibling nodes, not parent-child nodes.
 :::
 
-### `pan-intercept-direction`
+### `pan-intercept-direction` <AndroidOnly /> <IOSOnly />
 
 ```ts
 // DefaultValue: 2 /* none */
@@ -396,7 +396,7 @@ For example, when `pan-intercept-direction` is set to `0`, indicating a horizont
 On the Android platform, this setting only applies to `view`, `scroll-view`, and `list` elements, and the corresponding Android View needs to be created for the node to take effect. It may also require setting [`flatten`](#flatten) to `false` for it to work.
 :::
 
-### `pan-intercept-scope`
+### `pan-intercept-scope` <AndroidOnly /> <IOSOnly />
 
 ```ts
 // DefaultValue: 6 /* none */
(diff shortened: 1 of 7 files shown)


toretto-wt (approved):

Huxpro (approved):
Great!

Huxpro (changes requested):
One compatibility gap remains before this is safe to merge: this PR newly exposes `enableEventThrough` as a public page API/configuration option, but it adds neither availability/version metadata nor a compatibility badge for it. The compatibility-data change only covers the three new iOS pan-interception attributes in `elements/view.json`; there is no corresponding metadata or explicit “since” information for `enableEventThrough`.

Please document its supported platforms and minimum Lynx versions (through compat data if this API category supports it, otherwise with an explicit compatibility/version callout in both English and Chinese). This is especially important because the surrounding text presents it as part of the behavioral contract for `event-through`.

The iOS-only attributes themselves are handled well: they have platform badges, iOS 4.1 metadata, and bilingual API-reference coverage. The duplicated behavior explanation between `<page>` and `<view>` is also aligned across English and Chinese.

Huxpro (dismissed):
The iOS `hitTest:` integration example is reversed and would preserve hits inside `_lynxView` while returning `nil` for a valid non-Lynx sibling result—the opposite of the surrounding guidance. To exclude the Lynx page and allow another platform view to receive the touch, the parent should return `nil` when `result` is `_lynxView` or one of its descendants, and otherwise preserve `result` (for example, return `result` when it is nil or is *not* a descendant, then return nil). Please fix this in both locales before merging.

Please also make the release boundary explicit for the newly documented iOS pan-intercept properties. The upstream change added the declarations in `@lynx-js/types` 4.1.1 while annotating the SDK behavior as `@since 4.1`; the docs should distinguish the runtime compatibility floor from the first package version containing the TypeScript props rather than exposing only a 4.1 SDK badge.

rAY-ooo:
event-through在ios端适配的示例是对的，另外，那个版本也是4.1，这里还需要确认一下具体的小版本号吗？

Huxpro (changes requested):
Thanks for the clarification on the iOS wrapper topology. I re-evaluated it and dismissed my earlier review: the `hitTest:` example is correct when the shown parent is the Lynx wrapper and the intended responder is that wrapper’s platform sibling.

One version-metadata blocker remains. The implementation commit for these three filtered pan-intercept props explicitly bumped the exported type packages to `4.1.1`, while the declarations themselves annotate the runtime capability as `@since 4.1`. Please keep the iOS SDK compatibility entries at 4.1, but add a brief bilingual note that the TypeScript properties require `@lynx-js/types` 4.1.1 or later. The minor package boundary matters because developers on a 4.1 runtime with an older types package otherwise will not have these newly documented props in their public typings.

Huxpro (changes requested):
Two release/behavior metadata issues remain on the updated head:

1. The new bilingual note says these props require `@lynx-js/types` 4.1.1+, but 4.1.1 is not currently published; npm still exposes 4.1.0 as latest. Upstream lynx#7558 does bump the package source to 4.1.1 while adding these declarations, so the intended future boundary is clear, but the website should not present an unavailable package version as an installable requirement. Please either wait for and verify the actual publication, or explicitly mark the TypeScript declarations as unreleased/currently unavailable until that happens.

2. `ios-pan-intercept-view-tag` is documented with `DefaultValue: 0`, while the upstream public declaration uses `@defaultValue undefined`, and the implementation separately tracks whether the tag property was supplied. The prose in this PR also correctly says an omitted property does not match a target whose tag is 0. Please change the displayed default to `undefined` in both English and Chinese so the API signature and behavior are consistent.

The current Netlify preview is also failing for this head. These points should be resolved before merge.

Huxpro (changes requested):
This revision resolves the earlier `ios-pan-intercept-view-tag` default mismatch, but it is not yet safe to merge:

1. **The new `<page>` compatibility summary silently renders nothing.** `APISummary` requires either an explicit `query` prop or an `api` frontmatter value. `page.mdx` adds a bare `<APISummary />` without frontmatter, and the new `elements/page.json` contains only the nested `enableEventThrough` statement rather than a root `<page>` compatibility statement. Use an explicit query such as `elements/page.enableEventThrough`, or add the appropriate page-level metadata and root compatibility data. Ensure the visible summary and table point to the same source.

2. **The typings-availability note is incorrect/outdated.** The page now says the three iOS filtering properties are “not yet published in `@lynx-js/types`,” but the official types changelog records the pan-intercept properties in `@lynx-js/types` 4.1.0, and the public declarations contain the three `ios-pan-intercept-*` properties with a Lynx 4.1 runtime annotation. Replace the note with the exact package floor (currently 4.1.0+) and keep that distinct from the iOS SDK 4.1 compatibility badge.

3. **The diff includes substantial unrelated API-documentation drift.** In addition to the event-through and pan-interception work described by the PR, it rewrites `boundingClientRect`/SelectorQuery documentation and modifies the spelling dictionary. Those changes need to be split into a focused PR or separately justified and audited; they materially expand this change beyond a low-risk compatibility clarification.

4. **Keep the duplicate `enableEventThrough` descriptions aligned.** This adds a page-element reference while #1269 introduces another generated Lynx Config field reference. Please establish which surface is canonical or ensure the wording, platform/version metadata, defaults, and links are synchronized before both references land.

The branch is also currently non-mergeable. Please resolve the content issues and rebase after the current `main` updates.

Huxpro (changes requested):
Thanks for addressing the previous blockers: the compatibility query is now explicit, the iOS filtering properties include their `@lynx-js/types` 4.1.0 boundary, and the unrelated SelectorQuery expansion has been removed.

Two substantive issues still block merging:

1. **`enableEventThrough` is categorized and documented as a `<page>` element API, but it is a template configuration property.** This revision adds an `elements/page.enableEventThrough` compatibility entry, APISummary/APITable output, and a property-like section under `<page>`. At the same time, #1269 adds the generated canonical `@lynx-js/type-config.Config.enableEventThrough` reference with essentially the same signature, description, default, platforms, and LynxSDK floor. Landing both creates two independently maintained references and also counts a config key as an element API. Please establish one canonical config reference, link to it from the `<page>`/`event-through` explanation, and keep compatibility data in the correct config/feature taxonomy. Whichever user-facing reference remains must also state the package boundary: `enableEventThrough` is available in `@lynx-js/type-config` starting with 4.1.0, independently of the LynxSDK 3.2 runtime floor.

2. **The iOS `hitTest` workaround is still ambiguous about the native view hierarchy.** The prose says a sibling node should receive the event, while the sample returns the result only when it is `nil` or a descendant of `_lynxView`, and returns `nil` for every other descendant found by `super`. Whether that passes the touch onward or discards the intended responder depends on whether the target is a sibling of `LynxView`, a sibling of the overridden parent container, or elsewhere. Please document the exact tested hierarchy, identify which container overrides `hitTest`, and explain why the shown return values allow the intended target to receive the touch. Keep the English and Chinese guidance aligned.

Because these affect API taxonomy, version metadata, duplicate-reference ownership, and correctness of the integration sample, I cannot treat the change as low-risk yet.

Huxpro (changes requested):
Thanks for moving `enableEventThrough` into the feature/config taxonomy and removing the duplicate `<page>` API surface. Two substantive blockers remain on this head:

1. **The new canonical reference is currently broken and not version-aware.** Both the English/Chinese prose and `features/enable-event-through.json` point to `api/rspeedy/type-config.config.enableeventthrough`, but that page does not exist on `main`; it exists only in the still-unmerged #1269. The prose additionally uses absolute unversioned `https://lynxjs.org[/zh]/api/...` URLs, which target the stable-root site rather than the documentation version that contains the referring page. Please coordinate the canonical Config reference first, use a branch-relative/version-aware destination, and verify that the destination visibly carries the platform, LynxSDK 3.2, and `@lynx-js/type-config` 4.1.0 boundaries. Otherwise this PR lands compatibility metadata whose user-facing link is a 404 or points at the wrong site version.

2. **The iOS `hitTest` integration remains hierarchy-dependent and ambiguous.** The sample returns `nil` for every non-Lynx descendant returned by `[super hitTest:withEvent:]`. That is correct only when the intended responder is outside the overridden wrapper (for example, a sibling of the wrapper); it discards a valid target that is a sibling of `LynxView` inside that wrapper. Please state the exact tested native view hierarchy and which container overrides `hitTest`, or revise the sample to preserve the intended non-Lynx result. Keep the English and Chinese guidance aligned.

The iOS pan-intercept package/runtime metadata and the new feature taxonomy are now aligned, but the broken cross-site reference and unresolved integration ambiguity prevent a low-risk merge.

Huxpro (changes requested):
Re-reviewed the current head after #1269 merged. That resolves the earlier 404 dependency, and the three new iOS pan-intercept props now look internally aligned (`@lynx-js/types` 4.1.0+, iOS SDK 4.1 compatibility data, iOS-only badges, and `undefined` defaults). I still don't think this is safe to merge yet:

1. **The canonical `enableEventThrough` reference is still not aligned with the new `<view>` documentation.** This PR says the config property requires `@lynx-js/type-config` 4.1.0+, independently of the LynxSDK 3.2 runtime floor. The canonical generated `type-config.config.enableeventthrough.md` page on `main` currently exposes platforms, `Since: LynxSDK 3.2`, and the default, but does not expose that package-version boundary. Please update the canonical source (upstream TSDoc + regenerated reference, if needed) so the package floor is visible there too rather than maintaining a stronger version contract only on the element page. The links in this PR should also be branch/version-relative instead of absolute `https://lynxjs.org[/zh]/api/...` links, so versioned docs do not jump to a different stable-root API surface.

2. **The iOS `hitTest:` integration remains hierarchy-dependent but the prose still only says “sibling node.”** The sample is correct for the topology discussed earlier only when the overridden container is the Lynx wrapper and the intended responder is a sibling *of that wrapper*. If the intended responder is a sibling of `LynxView` inside the overridden parent, `[super hitTest:]` can return that valid non-Lynx descendant and this sample then replaces it with `nil`. Please state the exact tested hierarchy and which native container overrides `hitTest:` in both English and Chinese (or adjust the sample to preserve the intended non-Lynx result).

CI/deploy status is green on this head, but these are API-version/cross-site-consistency and integration-correctness issues, so I’m keeping it blocked under the low-risk merge bar.

Huxpro (changes requested):
The latest commit resolves the two blockers from my previous pass: the `Config.enableEventThrough` link is now version-relative, the canonical page is updated with the `@lynx-js/type-config` 4.1.0 package floor, and the iOS `hitTest:` guidance now states the exact wrapper/sibling topology and warns about the different inner-sibling case.

One source-of-truth blocker remains before this is safe to merge: this PR updates `docs/{en,zh}/api/rspeedy/type-config.config.enableeventthrough.md` directly. The repository's API-authoring guide explicitly says these TypeDoc-generated `docs/.../api/...` files must not be edited manually because regeneration overwrites them; the source comments must be changed and the API reference regenerated instead. Since the package-version note is important compatibility metadata, landing it only in the generated output would make the canonical reference drift again on the next TypeDoc run.

Please put the `@lynx-js/type-config` 4.1.0 availability boundary in the actual TypeScript/TSDoc source used by TypeDoc (upstream if that source lives outside this repo), then regenerate the EN/ZH pages. If the generator cannot currently express package-version metadata, that needs a deliberate durable mechanism rather than a hand edit.

The current head's GitHub Actions `CI` and `Compat Data Stats` runs are also still in `action_required` with no jobs started, so they have not yet provided the normal validation signal. Netlify checks are green, but I would not merge until the source-of-truth issue is fixed and the required workflows have actually run.

Huxpro (dismissed):
The latest commit addresses the previous source-of-truth design issue in the right direction: the package-version boundary is now generated from upstream `typePackageSince` metadata instead of being maintained only by hand in the website output, and the EN/ZH references are described as regenerated from that source.

One substantive dependency remains before this is safe to merge: the upstream source-of-truth change, lynx-family/lynx#8483, is still open and unmerged. That means this PR's generated `Config.enableEventThrough` pages are currently ahead of the canonical Lynx config source and cannot yet be reproduced from the merged upstream branch. Please land #8483 first, then regenerate/rebase this website PR from the merged source (or otherwise establish an equivalent merged canonical source) so the compatibility metadata cannot disappear on the next TypeDoc regeneration.

Also, the `CI` and `Compat Data Stats` workflow runs for this exact head completed with `action_required` and did not execute jobs. Netlify/CLA/CodeRabbit are green, but the normal repository validation signal is still missing. Please get the required workflows to actually run before merge.

rAY-ooo:
@pilipala195 帮忙看看lynx config这个需要添加typePackageSince字段吗？感觉有点奇怪，理论上config应该是跟随sdk version的

rAY-ooo:
@Huxpro 帮忙判断一下是否要在config中添加这个字段～

pilipala195:
LGTM

rAY-ooo:
@Huxpro 请求关注🙏

Huxpro (changes requested):
Codex-assisted review of the exact `93ab9ff` head. The rebase/squash resolves the stale generated-file diff, but it also regresses two compatibility-version boundaries that were present in the previously reviewed state:

1. The three newly documented `ios-pan-intercept-*` properties now expose only the iOS SDK 4.1 floor. The merged `@lynx-js/types` source/changelog shows these declarations first ship in `@lynx-js/types` 4.1.0, and the previous head explicitly distinguished that package floor from the runtime floor. Please restore that bilingual package-availability note (`@lynx-js/types` 4.1.0+) while keeping the iOS SDK compatibility data at 4.1.

2. The new `features/enable-event-through.json` entry links the canonical `Config.enableEventThrough` reference and records the LynxSDK 3.2 runtime floor, but the canonical generated page on current `main` still does not expose the independent `@lynx-js/type-config` 4.1.0 package floor. The durable upstream `typePackageSince` change in `lynx-family/lynx#8483` is still open/unmerged. Please land that source-of-truth change (or an equivalent durable mechanism) and regenerate the canonical reference rather than hand-editing generated output.

These are compatibility-metadata regressions on APIs this PR newly documents/references, not broader site-wide follow-ups. The exact head's `CI` and `Compat Data Stats` runs also completed as `action_required` without executing jobs, so the normal target-branch validation signal is still missing; please get those checks to actually run before merge.

This review is Codex-assisted. Please contact @Huxpro in person for a second opinion.

rAY-ooo:
@Huxpro what should i do

PR #1165, opened 2026-07-10
docs: add Lynx 3.9 release blog

WATER1350:
## Summary
- Add Lynx 3.9 blog entries in Chinese and English and register them in blog meta files.
- Replace the initial performance-report draft with release announcement content for Lynx 3.9.
- Cover A2UI, Autolink, ReactLynx, Element, CSS, API, DevTool, and the HarmonyOS image performance note.

## Validation
- pnpm exec prettier --check docs/zh/blog/lynx-3-9.mdx docs/en/blog/lynx-3-9.mdx docs/zh/blog/_meta.json docs/en/blog/_meta.json
- pnpm exec cspell docs/zh/blog/lynx-3-9.mdx docs/en/blog/lynx-3-9.mdx
- pnpm exec node scripts/ci/check-asset-urls.mjs docs/zh/blog/lynx-3-9.mdx docs/en/blog/lynx-3-9.mdx docs/zh/blog/_meta.json docs/en/blog/_meta.json
- pnpm build

- Publish Chinese and English Lynx 3.9 release blog posts.
- Register the posts in blog metadata.
- Document Lynx 3.9 CSS behavior for `flex-basis` and `enableNewSticky`.
- Cover Autolink, ReactLynx, Element, Lynx API, DevTool, and HarmonyOS image performance updates.
- Add Yiyi Lin to author metadata.
- Update the `@lynx-example/css-api` dependency to `0.10.2`.
- Add project-specific terms to the cspell dictionary.

PR #1237, opened 2026-07-31
docs(blog): 3.9 blog review follow-ups

Huxpro:
Stacked on top of #1165 (`p/linyiyi/add_3_9_blog`).

## Summary

- Remove the **Lynx for AI** subsection from both `docs/zh/blog/lynx-3-9.mdx` and `docs/en/blog/lynx-3-9.mdx`, and drop the corresponding mention from the intro paragraph.
- Deep-link the `<frame>` props to their exact anchors on the docs site instead of only linking the element page: `#enable-multi-async-thread`, `#preset-height`, `#preset-width`, `#bindloadmetrics`.
- Remove the `<image>` `region-to-decode` and `<overlay>` `mode` notes — see below.
- Fix the zh `!important` link: the anchor on the Chinese selectors page is `#important-支持`, not `#important-support` (the previous link scrolled nowhere).
- Retitle the CSS section to lead with the "closer to the Web" framing:
  - zh: `## CSS 能力` → `## 更贴近 Web 的 CSS`
  - en: `## CSS Updates` → `## CSS, Closer to the Web`
- Rewrite the HarmonyOS image performance note to describe the new native image pipeline, and gloss the two metrics on first use — 总耗时 / total time (`TotalTime`) and CPU 耗时 / CPU time (`CPUTime`) — instead of leaving bare identifiers plus the internal `image_animation` test-case name.
- Enable the web preview for the CSS `position` example: bump `@lynx-example/css-api` to `0.10.2` and set `defaultTab="web"`.

## Why the image and overlay notes were removed

Both described capabilities the 3.9 docs do not back up, so there was nothing correct to link to:

- **`<image>` `region-to-decode`** does not appear on the 3.9 image API page, nor anywhere in this repo — the blog draft was its only mention.
- **`<overlay>` `mode`** exists, but is documented with `<IOSOnly />` and describes `window`/`top`/`page`/`others` for iOS, contradicting the draft's HarmonyOS `page`/`top` framing.

If either capability did ship in 3.9, the docs need updating first; the notes can then come back with working links.

## Verification

Every link in the blog was checked against the **released 3.9 docs** (`https://lynxjs.org/3.9/...`), not just local `main`:

- All internal doc pages referenced by the blog return `200` on the 3.9 site.
- All four `<frame>` anchors (`enable-multi-async-thread`, `preset-height`, `preset-width`, `bindloadmetrics`) are present in the rendered 3.9 HTML.

For the web preview:

- `@lynx-example/css-api@0.10.2` is the first release containing `dist/position.web.bundle` — `0.10.1` predates lynx-family/lynx-examples#386 and shipped only the lynx bundle. Verified by inspecting both tarballs.
- After the bump, `docs/public/lynx-examples/css-api/example-metadata.json` resolves the `position` entry with `"webFile": "dist/position.web.bundle"`, and the bundle is present on disk.

## Checks

- `pnpm exec prettier --check` — pass
- `pnpm exec cspell` — 0 issues

### DIFF
--- a/docs/en/blog/lynx-3-9.mdx
+++ b/docs/en/blog/lynx-3-9.mdx
@@ -107,7 +107,7 @@ Lynx API [`requestResourcePrefetch`](/api/lynx-api/lynx/lynx-request-resource-pr
 
 ### Performance
 
-On HarmonyOS, the new image strategy optimization improves scenarios such as `image_animation`, reducing `TotalTime` and `CPUTime` by roughly 10%.
+On HarmonyOS, the new native image pipeline is enabled by default. It moves image loading, decode caching, and animated image frame scheduling down into the image node, which reduces per-frame computation and refresh overhead. In image animation benchmarks, both total time (`TotalTime`) and CPU time (`CPUTime`) dropped by roughly 10%.
 
 ## Upgrade Guide
 
--- a/docs/en/blog/lynx-3-9.mdx
+++ b/docs/en/blog/lynx-3-9.mdx
@@ -71,6 +71,7 @@ Lynx 3.9 also adds the following CSS capabilities:
   img="https://lf-lynx.tiktok-cdns.com/obj/lynx-artifacts-oss-sg/lynx-website/assets/doc/css/position.jpg"
   entry="src/position"
   defaultEntryFile="dist/position.lynx.bundle"
+  defaultTab="web"
 />
 
 ## ReactLynx
--- a/docs/en/blog/lynx-3-9.mdx
+++ b/docs/en/blog/lynx-3-9.mdx
@@ -17,7 +17,7 @@ Lynx 3.9 is now officially released!
 
 This release focuses on three themes: [Autolink](/guide/autolink) for simpler native capability integration, CSS behavior that is closer to the Web platform, and ReactLynx updates around Portal, rendering, and bundle output.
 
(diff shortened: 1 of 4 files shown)


Huxpro (commented):
The follow-up is directionally good: the corrected anchors, removal of the unsupported Lynx-for-AI section, and the `css-api@0.10.2` web preview are useful improvements. I would still keep this stacked PR and its parent draft unmerged until the two explicitly acknowledged API-reference gaps are resolved:

1. **`<image> region-to-decode` is currently advertised as a Lynx 3.9 API but has no API-reference entry anywhere in this repository.** Add the bilingual API documentation, supported-platform badges, minimum-version/compatibility metadata, and usage semantics, or remove it from the release blog.
2. **`<overlay> mode` currently contradicts the global API reference.** The blog says HarmonyOS supports `page`/`top`, while the reference marks `mode` as iOS-only and describes `window`/`top`/`page`/custom class names. Update the API reference and compatibility data/badges to reflect the actual 3.9 HarmonyOS contract, or remove/correct the HarmonyOS claim in the blog.

The strengthened HarmonyOS image-performance paragraph also introduces a broader public claim (“enabled by default” and roughly 10% lower `TotalTime`/`CPUTime`) that is not documented elsewhere on the site. Please identify the exact 3.9 component/package/platform scope and link a reproducible benchmark or canonical performance note so this does not become an isolated release-blog claim that later drifts from the global docs.

CI is green on this head, but the PR is still a draft and the Cloudflare preview was still in progress at review time.

PR #1238, opened 2026-07-31
docs(blog): rewrite the 3.9 ReactLynx section

upupming:
Rewrites the ReactLynx section of the 3.9 blog, in both zh and en.

- Split into `createPortal` / Snapshot / Minimizer subsections
- Add `createPortal` usage, capabilities and limitations
- Explain slots and wrapper elements before using the terms, and show the compiled Snapshot output before/after
- Recommend `OverlayView` over the raw `<overlay>` element
- Use `@lynx-js/react` versions (0.121.0 / 0.120.0) instead of the engine version

### DIFF
--- a/docs/en/blog/lynx-3-9.mdx
+++ b/docs/en/blog/lynx-3-9.mdx
@@ -76,9 +76,42 @@ Lynx 3.9 also adds the following CSS capabilities:
 
 ## ReactLynx
 
-ReactLynx first adds [`createPortal`](/api/react/Function.createPortal) in 3.9. Portal lets a vnode tree continue rendering into a specified container node outside the current component hierarchy, which is useful for overlays, floating layers, and UI that needs to mount across containers.
+ReactLynx brings `createPortal` and an optimization that reduces both the depth and the number of Snapshots, plus a dual-thread Minimizer on the build side.
 
-Previously, this kind of cross-container rendering often required extra wrapper nodes, host-side conventions, or scattered mounting logic. Now `createPortal` brings the pattern into the ReactLynx API surface, so components can keep following React data flow while the final render location is delegated to the target container.
+### `createPortal`
+
+[`createPortal`](/api/react/Function.createPortal) renders a component subtree into a container outside the current component hierarchy. It suits overlays, dialogs, and other UI that needs to escape the surrounding layout: components stay organized around React data flow, while the final render location is delegated to the target container. The API is available from `@lynx-js/react` 0.121.0.
+
+#### Usage
+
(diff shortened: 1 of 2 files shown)


Huxpro (changes requested):
The rewrite is substantially clearer, and using `@lynx-js/react` package versions is the right direction, but this should not merge while the canonical API documentation remains behind the blog.

Blocking issues:

1. The linked `Function.createPortal` API page currently renders an incorrect return type (`useErrorBoundary<any> | null`) instead of the portal vnode return type. Please fix the generated source/API overlay rather than relying on the blog to compensate.
2. The blog now states that `createPortal` is available from `@lynx-js/react` 0.121.0, but the API page has no availability/version metadata or badge. The version boundary should be visible on the canonical reference page as well.
3. Important behavioral contract is documented only here: actual-tree event bubbling, background-thread-only rendering/no main-thread first-screen participation, and no cross-page or cross-native-container mounting. These are API semantics and limitations, so they need to be added to `Function.createPortal` in both English and Chinese and kept aligned with this post.
4. The Snapshot optimization is stated as enabled by default in 0.120.0, but there is no linked canonical documentation for the optimization/version boundary. Please add or link a stable reference rather than making the release blog the sole source of truth.
5. The production-page measurements (175→123 Snapshots and 76→25 wrapper calls) need enough scope to be interpretable: identify the measured page/workload, versions/build configuration, and whether these are representative or a single case.

Because this PR is stacked on #1165, it should also remain blocked until the base release post's outstanding API-reference and compatibility issues are resolved.

Huxpro:
The five blocking items are now resolved:

1. **API page return type**: fixed by lynx-family/lynx-website#1240 (merged) + lynx-family/lynx-stack#3348 (merged; ships in `@lynx-js/react` 0.123.3 per lynx-family/lynx-stack#3350).
2. **Availability metadata on the canonical page**: the API page now carries an Availability block (added in 0.121.0; `ReactNode` return type since 0.123.3, with the pre-fix signature called out).
3. **Behavioral contract in the API page (en+zh)**: present in both locales and sourced from the shipped TSDoc.
4. **Snapshot optimization version boundary**: the `0.120.0` mention now links the `@lynx-js/react` changelog anchor (`CHANGELOG.md#01200`) as the stable reference (ab122887).
5. **Measurement scope**: the 175→123 / 76→25 numbers are now framed as an illustrative single-production-page measurement with per-page variance noted (ab122887). The exact workload identity is only known to the original measurer; the wording no longer presents the numbers as representative.

Huxpro (approved):

PR #1246, opened 2026-08-02
feat(blog): add author Group concept and ReactLynx Team

Huxpro:
## Overview

A blog post's `authors` frontmatter can now name a **group** as well as an individual. A group is a name plus an array of author ids, and it renders as a single entry — stacked avatars, the group name, and the member count — instead of expanding into one row per person.

`src/components/blog-avatar/groups.json`:

```json
{
  "id": "reactlynx",
  "name": "ReactLynx Team",
  "name_zh": "ReactLynx 团队",
  "members": ["upupming", "HuJean", "huxpro", "colinaaa", "gaoachao", "hzy", "Yradex"]
}
```

Used the same way as an individual: `authors: ['reactlynx']`, or mixed — `authors: ['liushouqun', 'reactlynx']`.

## Rendering

**Post header (full mode)** — up to four stacked avatars, then a `+N` bubble, followed by the group name and `7 members` / `7 位成员`. Hovering opens a card with every member's avatar and full name list.

**Blog index card (compact mode)** — stacked avatars plus the group name. Stacking is now capped at four for plain author lists too, so long author lists no longer push the names out of the card.

Members are sorted alphabetically at render time by their display name in the active locale (`Intl.Collator`), so the source order in `groups.json` doesn't matter and `zh` gets pinyin ordering.

Unknown ids are dropped, as before; a group whose members all fail to resolve is dropped too.

## Author entries added

`hzy` (Zhiyuan Hong) and `Yradex` (Ziqi) were not in `authors.json` and are needed for the ReactLynx Team roster.

Two placeholders to correct before this leaves draft:

- `name_zh` for both is set to the English name — the Chinese names weren't available to confirm, and guessing at someone's characters seemed worse than falling back.
- `Yradex` is listed as `Ziqi` only; the full name is unknown.

## Verification

Typechecked (`tsc --noEmit` — no new errors in `blog-avatar`) and rendered against the dev server: post header, hover card, and blog index all verified by screenshot.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CzfaaQBfP5TyQHePBcdWQx)_

- Support author groups in blog frontmatter with localized names and member IDs.
- Render groups with stacked avatars, member counts, hover cards, and compact layouts.
- Sort localized group members, exclude unknown or duplicate authors, and cap avatar stacks at four.
- Add ReactLynx Team membership and missing author records.
- Apply theme-safe avatar masking and background styling.

### DIFF
--- a/src/components/blog-avatar/authors.json
+++ b/src/components/blog-avatar/authors.json
@@ -406,6 +406,34 @@
     }
   },
   {
+    "id": "hzy",
+    "name": "Zhiyuan Hong",
+    "name_zh": "Zhiyuan Hong",
+    "title": "Engineering @ Lynx",
+    "title_zh": "工程师 @ Lynx",
+    "image": "https://github.com/hzy.png",
+    "socials": {
+      "github": {
+        "username": "hzy",
+        "link": "https://github.com/hzy"
+      }
+    }
+  },
+  {
+    "id": "Yradex",
+    "name": "Ziqi",
+    "name_zh": "Ziqi",
+    "title": "Engineering @ Lynx",
+    "title_zh": "工程师 @ Lynx",
+    "image": "https://github.com/Yradex.png",
+    "socials": {
+      "github": {
+        "username": "Yradex",
+        "link": "https://github.com/Yradex"
+      }
+    }
+  },
+  {
     "id": "Neilcc",
     "name": "Chengcheng Zhu",
     "name_zh": "朱承丞",
--- a/src/components/blog-avatar/groups.json
+++ b/src/components/blog-avatar/groups.json
@@ -1,0 +1,16 @@
+[
+  {
+    "id": "reactlynx",
+    "name": "ReactLynx Team",
+    "name_zh": "ReactLynx 团队",
+    "members": [
+      "upupming",
+      "HuJean",
+      "huxpro",
+      "colinaaa",
+      "gaoachao",
+      "hzy",
+      "Yradex"
+    ]
+  }
+]
--- a/src/components/blog-avatar/index.module.less
+++ b/src/components/blog-avatar/index.module.less
@@ -82,20 +82,16 @@
   }
 }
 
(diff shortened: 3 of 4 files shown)


Huxpro (commented):
**Blocking review:**

- The group-member UI is exposed only through `Popover trigger="hover"` wrapped around a non-focusable `<span>`. Keyboard users cannot open it, and hover-only disclosure is unreliable on touch devices. Please use a focusable semantic trigger (or equivalent keyboard/focus handling), expose the group name/member count with appropriate accessible labeling, and make the member list reachable without requiring pointer hover.
- The two new author records are still explicitly placeholder metadata: `name_zh` falls back to English for both, and `Yradex` has only the partial display name `Ziqi`. Please confirm the intended public names with the authors before merging rather than publishing uncertain identity data.
- Since the deduplication behavior is now relied on by stacked blog PR #1247, please add focused coverage for mixed individual/group bylines such as `['Yradex', 'reactlynx']`, unknown members, and an empty resolved group. This logic changes attribution, so a regression would be visible and substantive.

No API version or compatibility metadata is affected by this UI-only change, but the accessibility and attribution issues make it unsafe to merge as-is.

Huxpro:
New concrete blocker from the first consumer of this feature: stacked PR #1247 fails the production build when `authors: ['Yradex', 'reactlynx']` actually renders a group. The SSG-MD build crashes in Semi UI's tooltip/popover lifecycle with `ReferenceError: window is not defined` while rendering `/zh/blog/reactlynx-instant-first-frame-rendering` (and another page using `BlogHeader`).

#1246's own CI passes because no published page exercises a group entry, so this path is currently untested. Please make the group disclosure SSR-safe (for example, avoid mounting the browser-only Popover during SSG or isolate it behind an appropriate client-only boundary) and add a build/SSR regression fixture that renders at least one group. This is in addition to the accessibility, author-metadata, and attribution-test blockers already raised.

PR #1243, opened 2026-08-01
docs(blog): tighten the Autolink opening into one paragraph

Huxpro:
Stacked on `p/linyiyi/add_3_9_blog` (#1165), following the same pattern as #1238.

## What changed

The Autolink section opened with two paragraphs — a problem statement, then the definition. Review feedback was that the problem statement ran long before getting to the point. Folded into one paragraph: a short motivation, then the definition of a native library and Autolink, rephrased to read as a single thought.

**Before** (two paragraphs):

> 在 3.9 之前，把一段原生能力接进 Lynx 应用，需要应用侧自己写注册代码：元件、原生模块和 Service 各有各的注册入口，Android 和 iOS 还要各写一遍。想把同一份能力用到第二个应用上，只能把接入步骤写进 README，让对方照着抄一遍。缺的不是某一个 API，而是「原生库」这个可以被发布、被安装的单位。
>
> Lynx 3.9 定义了这个单位。一个 Lynx 原生库就是一个 npm 包，在包根目录用 `lynx.lib.json` 声明自己的原生入口；Autolink 则是在构建期发现这些库、并把它们接进宿主应用的机制。

**After** (one paragraph):

> 在 3.9 之前，一段原生能力只能在每个应用里各接一次：元件、原生模块和 Service 各有各的注册入口，Android 和 iOS 还要各写一遍，复用到下一个应用往往意味着照着 README 再抄一遍。Lynx 3.9 把这件事收敛成一个可以发布、可以安装的单位：一个 Lynx 原生库就是一个 npm 包，在包根目录用 `lynx.lib.json` 声明原生入口；Autolink 会在构建期发现这些库，并把它们提供的能力接进宿主应用。

Dropped the two editorializing sentences (「缺的不是某一个 API…」 and 「Lynx 3.9 定义了这个单位」) — the same framing survives in 「收敛成一个可以发布、可以安装的单位」, and the section still closes on 「把原生能力做成库发出去、和在另一个应用里装上它，是同一条路径的两端」.

Every link in the original two paragraphs is preserved. The rest of the section — 使用库 / 开发库 and the build-time mechanism — is untouched.

Same edit applied to the English blog.

## Checks

- `prettier --check` — pass
- `cspell` — 0 issues
- `node scripts/ci/check-asset-urls.mjs` — pass

---
_Generated by [Claude Code](https://claude.ai/code/session_01FVdDv5kxo9CgBZqTNuQeY7)_

PR #1249, opened 2026-08-02
docs: close the 3.9 release-blog review blockers

Huxpro:
Stacked on `p/linyiyi/add_3_9_blog`, addressing the four blockers in the [CHANGES_REQUESTED review on #1165](https://github.com/lynx-family/lynx-website/pull/1165#pullrequestreview-4838246978).

## 1. Portal limitation said "containers" where it meant native containers

Correct catch — the shortened wording was ambiguous against the section's own usage example, since every portal mounts into *some* element container.

> zh: 不支持跨页面、跨容器挂载 → **不支持跨页面或跨原生容器挂载**
> en: across pages or containers → **across pages or across native containers**

Now matches the `createPortal` contract exactly.

## 2. `flex` contradicted the linked canonical page

The blog was the only place the `0` → `0%` change was written down. Both locales of `/api/css/properties/flex` now version it:

> …the shorthand expands to `flex: <flex-grow> 1 0%`. <VersionBadge>3.9</VersionBadge> Before 3.9 the omitted `flex-basis` expanded to a unitless `0`; browsers use `0%`, so this was aligned with the Web.

Applied to both the one-value and two-value expansions, plus the syntax comment. `VersionBadge` was not previously imported on these pages; it is now.

## 3. `enableNewSticky` was undocumented

`/api/css/properties/position` described only the legacy `scroll-view`-child behavior. Both locales now carry an info block on `#### sticky`: the 3.9 switch name, what standard sticky changes (normal-flow layout, offset against the nearest scrolling container, no longer requiring a direct `scroll-view` child), and that it is **off by default** so existing pages are unaffected.

The EN page also had a typo in that paragraph ("Is The element is positioned"), fixed in passing.

## 4. Performance figure was not reproducible

Removed the bare "roughly 10%" `TotalTime`/`CPUTime` claim rather than dressing it up. Without a published workload, device, build and sample size the number cannot be acted on, so the post now states the direction and tells readers to measure their own scenario.

> 在我们内部的图片动画基准场景中，总耗时（`TotalTime`）和 CPU 耗时（`CPUTime`）都有可观测的下降；具体幅度取决于图片规格、设备和页面结构，请以你自己场景的实测为准。

The Snapshot production counts (`175→123`, `76→25`) were already removed when that section was rewritten in #1243.

## Not addressed

**Dual-thread Minimizer availability metadata.** The review asks which `@lynx-js/react-rsbuild-plugin` / `@lynx-js/rspeedy` versions first support it. That is not derivable from this repository — the API pages document signatures only, and nothing here records the first supporting release. Rather than guess a version floor into a reference page, leaving it for someone with the release history. The blog paragraph makes no version claim, so nothing there is wrong today.

## Checks

- `prettier --check` — pass
- `cspell` — 0 issues

---
_Generated by [Claude Code](https://claude.ai/code/session_01FVdDv5kxo9CgBZqTNuQeY7)_

### DIFF
--- a/docs/en/api/css/properties/flex.mdx
+++ b/docs/en/api/css/properties/flex.mdx
@@ -3,7 +3,7 @@ title: flex
 api: css/properties/flex
 ---
 
-import { APISummary, APITable, Go } from '@lynx';
+import { APISummary, APITable, Go, VersionBadge } from '@lynx';
 
 <APISummary />
 ## Introduction
@@ -27,7 +27,7 @@ flex: none;
 flex: auto;
 
 /* One value, unitless number: flex-grow
-flex-basis is then equal to 0. */
+flex-basis is then equal to 0% (0 before Lynx 3.9). */
 flex: 2;
 
 /* One value, width/height: flex-basis */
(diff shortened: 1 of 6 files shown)


Huxpro (commented):
Blocking findings before merge:

1. **`enableNewSticky` needs per-platform version metadata, and `3.9` is not a safe global boundary.** The config plumbing and Android/iOS implementation were backported to both `release/3.8` and `release/3.9` (lynx#7009 / #7098 and #7003 / #7102), while the Harmony implementation I can find landed only on `develop` in lynx#7089. The generated config metadata currently says `since: 3.8` for Android, iOS, and HarmonyOS, which itself conflicts with that branch history. Please resolve the source-of-truth mismatch and document the exact first released version **per platform** with a compatibility badge/matrix. Do not say simply “From 3.9 onwards” until that is verified against released artifacts.

2. **The `flex-basis: 0%` boundary also needs its actual gate documented.** `enableFlexBasisZeroPercent` is recorded in the 3.9 branch metadata as `since: 3.8`, supported on Android/iOS/HarmonyOS, and defaulting to `false`. The proposed API text says the shorthand unconditionally changes in 3.9, but does not explain which 3.9 compiler/tooling version enables the flag, whether it is target-SDK-gated, or how older bundles behave on a newer runtime. Please document the real compatibility condition and first-supported tooling/runtime versions rather than using a bare `<VersionBadge>3.9</VersionBadge>`.

3. **The performance claim is still not reproducible.** Replacing “roughly 10%” with “came down measurably” removes the number but still makes an unsupported empirical claim. Either link a benchmark/report with workload, device, build and measurement details, or remove the benchmark-result claim entirely and describe only the architectural change.

The Portal wording correction is good. The parent release-blog PR also remains blocked on the dual-thread Minimizer package-version metadata that this PR explicitly leaves unresolved.

PR #1250, opened 2026-08-02
docs(blog): tighten the 3.9 post after a full read-through

Huxpro:
Stacked on `p/linyiyi/add_3_9_blog`. Nine edits from a read-through of the English post, mirrored in Chinese.

| # | Change |
|---|---|
| 0 | Restore the **roughly 10%** HarmonyOS figure |
| 1 | Fold "It also includes…" into the opening paragraph |
| 2 | `!important`: state the outcome **before** the snippet |
| 3 | Link `enableNewSticky` to its docs section |
| 4 | Gloss **Motion Path** with MDN, say "more Motion Path features" |
| 5 | Highlight the sticky lines in the `position` example |
| 6 | Retitle to **Portal**, link react.dev, restructure the intro |
| 7 | Portal notes: inline the link, drop the trailing sentence |
| 8 | Snapshot: keep the idea, defer the mechanics |
| 9 | Drop the odd "start with X, then evaluate Y" closing advice |

## Notes on a few of them

**0 — the 10% figure comes back.** The sentence already scopes it to our own benchmark, so the hedged rewrite was longer without being more honest. Reverted to `dropped by roughly 10%` / `均降低约 10%`.

**5 — sticky highlight.** `highlight="{139-142,166-168}"` covers the `stickyHeader` rule (`position: "sticky"`, `top`, `zIndex`) and the element that uses it, verified against `src/position/App.tsx`.

**6 — Portal.** Now leads with what a Portal is in React, linked to [react.dev](https://react.dev/reference/react-dom/createPortal), then when ReactLynx got one, then the single thing that differs:

> Since `@lynx-js/react` 0.121.0, Portal is available in ReactLynx through [`createPortal`](/api/react/Function.createPortal).
>
> Unlike React DOM, ReactLynx takes a [`NodesRef`](/api/lynx-api/nodes-ref) as the Portal container…

**7 —** now reads `React Context still flows across the Portal boundary… Three [notes](/api/react/Function.createPortal#remarks) differ from the Web:` with the link inside the sentence, and the standalone pointer after the bullets removed.

**8 — Snapshot** drops the diff-alignment explanation, which was the part that wanted a guide page rather than a release post:

> At compile time ReactLynx extracts the static element structure of your JSX into a **Snapshot**… (We will deep dive into Snapshots in a later post.)
>
> In `@lynx-js/react` 0.120.0 we extended Preact's diff algorithm so a child can declare which slot of its parent it belongs to. That lets the compiler drop the intermediate nodes it used to need, leaving a shallower element tree and fewer Snapshots — less work to render.

## Checks

- `prettier --check` — pass
- `cspell` — 0 issues
- `#sticky` resolves on both locales of the position page; highlight ranges verified against the example source.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FVdDv5kxo9CgBZqTNuQeY7)_

### DIFF
--- a/docs/en/blog/lynx-3-9.mdx
+++ b/docs/en/blog/lynx-3-9.mdx
@@ -71,7 +71,7 @@ Lynx 3.9 also adds the following CSS capabilities:
 
 - New [`grid-row`](/api/css/properties/grid-row) and [`grid-column`](/api/css/properties/grid-column) shorthands for setting the corresponding `grid-*-start` and `grid-*-end` values. For example, `grid-row: 1 / span 2` is equivalent to `grid-row-start: 1; grid-row-end: span 2`.
 - Standard `position: sticky` behavior, enabled through the [`enableNewSticky`](/api/css/properties/position#sticky) page switch.
-- HarmonyOS support for more [Motion Path](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_motion_path) features, including [`offset-path`](/api/css/properties/offset-path), [`offset-distance`](/api/css/properties/offset-distance), and [`offset-rotate`](/api/css/properties/offset-rotate).
+- HarmonyOS support for more [Motion Path](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Motion_path) features, including [`offset-path`](/api/css/properties/offset-path), [`offset-distance`](/api/css/properties/offset-distance), and [`offset-rotate`](/api/css/properties/offset-rotate).
 
 <Go
   example="css-api"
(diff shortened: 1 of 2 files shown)


Huxpro (commented):
Blocking notes before this stacked PR is merged:

1. Please do not restore the “roughly 10%” HarmonyOS performance claim without a reproducible source or enough methodology to evaluate it (device, image workload, build/config, sample size/runs, and how `TotalTime` / `CPUTime` were measured). Saying it came from an internal benchmark scopes the claim, but does not make the number verifiable. Keep the non-quantified architectural wording, link a benchmark report, or add the missing methodology.

2. The new `enableNewSticky` link points readers to the parent PR’s blanket 3.9 compatibility note, which is still unresolved. The switch/implementation history is platform- and branch-dependent, so the canonical section needs exact per-platform/version metadata or badges before the release post should direct readers to it as the source of truth.

3. This PR is stacked on #1165, which still has unresolved compatibility/version issues for the `flex-basis` behavior and the dual-thread Minimizer package floors. Please keep this unmerged until the base PR’s canonical API documentation accurately describes those boundaries.

The Portal rewrite and the links to the existing `createPortal` remarks look good.

PR #1252, opened 2026-08-03
[Doc] Document the desktop cover-view element

icecreamx10:
- Add English and Chinese API guides for the desktop overlay container.

- Explain when desktop custom native elements need cover-view for overlapping content.

- Record Clay macOS and Windows support starting with Lynx 3.9.

- Add the element to navigation and regenerate compatibility statistics.

TEST: API doc check, compatibility schema validation, CSpell, Prettier, and zhlint.

- Document the desktop-only `<cover-view>` element in English and Chinese.
- Explain when `<cover-view>` is required for overlapping native views.
- Document Windows child-window and macOS `ClayOverlayView`/`CALayer` behavior.
- Add API navigation entries and Clay Windows/macOS compatibility from Lynx 4.1.
- Regenerate API compatibility statistics.
- Correct UTF-8 conversion buffer sizing and failure handling.

### DIFF
--- a/docs/en/api/elements/built-in/_meta.json
+++ b/docs/en/api/elements/built-in/_meta.json
@@ -88,5 +88,10 @@
     "type": "file",
     "name": "title-bar-view",
     "label": "<title-bar-view>"
+  },
+  {
+    "type": "file",
+    "name": "cover-view",
+    "label": "<cover-view>"
   }
 ]
--- a/docs/en/api/elements/built-in/cover-view.mdx
+++ b/docs/en/api/elements/built-in/cover-view.mdx
@@ -1,0 +1,59 @@
+---
+api: elements/cover-view
+tag: 'XElement'
+---
+
+import { APISummary, APITable, ClayMacOSOnly, ClayWindowsOnly } from '@lynx';
+
+# `<cover-view>` <ClayWindowsOnly /> <ClayMacOSOnly />
+
+<APISummary />
+
+`<cover-view>` is a desktop-only container that renders its child subtree into a platform overlay surface above Clay's main rendering surface. Use it when Lynx content must cover an embedded native or platform view that would otherwise appear above normally rendered Lynx content.
+
+For overlays that only need to cover other Clay-rendered content, prefer a regular `<view>` with `position: fixed`.
+
+## Usage
+
+Position and size `<cover-view>` with the same CSS layout properties used by `<view>`. Its children are rendered into the platform overlay surface within those bounds.
+
+```tsx
+export function OverlayControls() {
+  return (
+    <cover-view className="cover">
+      <view className="controls">
+        <text>Controls rendered above native content</text>
+      </view>
+    </cover-view>
+  );
+}
+```
+
+```css
+.cover {
+  position: absolute;
+  top: 16px;
+  right: 16px;
+  width: 240px;
+  height: 48px;
+}
+
+.controls {
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0.72);
+}
+```
+
+## Platform behavior
+
+- On Windows, Clay renders the subtree into a native child overlay window and keeps that window positioned above the main rendering surface.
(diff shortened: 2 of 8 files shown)


Huxpro (changes requested):
Blocking on the compatibility boundary. The new compatibility data marks both Clay Windows and Clay macOS as available since Lynx 3.9, but the upstream implementations landed later: Windows in lynx-family/lynx@0d1852d and macOS in @cd7c53f. The Windows commit is not contained in release/3.9 or release/4.0; it is an ancestor of release/4.1. Please verify the macOS ancestry as well and replace the blanket 3.9 entries with the exact first-supported release for each desktop platform. Until those version floors are source-backed, the new badges and API page would publish incorrect compatibility metadata.

Huxpro (changes requested):
The new composition guidance and host-parenting examples are materially better, but the blocking compatibility issue is still unchanged on this head: `packages/lynx-compat-data/elements/cover-view.json` continues to publish both Clay Windows and Clay macOS as `version_added: "3.9"`. The Windows implementation is not present in release/3.9 or release/4.0 and first appears in the 4.1 line; macOS also needs an independently verified first-supported release. Please replace the blanket 3.9 entries with source-backed per-platform version floors before merging.

Also, the deploy preview for `7b09f7e` is currently failing, so the newly added pages and examples do not yet have a successful rendered preview. GitHub CI and compat-data checks pass, but that does not resolve the incorrect badges/version metadata.

icecreamx10:
@Huxpro Thanks for catching the compatibility boundary. I updated both `clay_windows` and `clay_macos` from `3.9` to `4.1` and regenerated `api-stats.json`.

I verified the upstream implementation commits against the release branches:

- Windows: [0d1852d](https://github.com/lynx-family/lynx/commit/0d1852dffe7ffdb85c3f9e989384a3b6f28580ae)
- macOS: [cd7c53f](https://github.com/lynx-family/lynx/commit/cd7c53fdc98f5454034b051a7979a5ce0c9eb826)

Neither commit is an ancestor of `release/4.0`; both are ancestors of `release/4.1`.

Validation completed locally:

- compat-data schema validation
- key-name lint
- stats regeneration
- compat-data Vitest suite (10/10)

Could you please take another look?

Huxpro (changes requested):
The 4.1 compatibility correction is now source-backed for both desktop platforms, and the latest preview is green. One substantive blocker remains in the Windows example in both the English and Chinese desktop guides.

`Utf8ToWide()` calls `MultiByteToWideChar(..., -1, ...)`, so the returned `size` includes the terminating NUL. It then allocates `std::wstring(size - 1, ...)` but passes `size` as the destination capacity on the second call. That writes `size` wide characters into storage sized for only `size - 1`, causing a one-code-unit buffer overrun in copy-pastable native code.

Please allocate `size` code units and then remove/resize past the trailing NUL, or call the conversion with an explicit source length and allocate exactly that returned length. Apply the same correction to both locale copies. I would not merge a native C++ guide with this memory-safety bug.

icecreamx10:
@Huxpro Thanks for catching this. Fixed in `35bbc244` in both locale copies.

`Utf8ToWide()` now allocates all `size` code units (including the terminating NUL), checks the second `MultiByteToWideChar()` result, and then resizes to `converted - 1` to remove the trailing NUL.

Validation completed:

- Prettier, CSpell, zhlint, API doc check, and `git diff --check`
- Extracted the shared, macOS, and Windows Native UI code blocks and verified that the test sources exactly match the documentation
- Rebuilt `libLynx.dylib` in the adjacent template-assembler checkout
- Compiled and ran the macOS example against a real Clay `LynxView`; verified native-window lookup, `NSTextField` attachment/frame, detach, and release
- Compiled and linked the complete Windows example against the real Lynx headers/library, then executed it with a minimal Win32 ABI shim on macOS
- Ran an ASan conversion matrix covering null/empty input, ASCII, Chinese, emoji/surrogate pairs, and lengths 1–4096

The current host does not have MSVC/Windows SDK, so the Windows execution is not a native Windows PE run. Could you please take another look?

Huxpro (approved):

PR #1253, opened 2026-08-03
docs(lynxtron): add lynxBridge API reference pages

zsy-jason:
Document the new `lynxBridge` main-process singleton exported from `@lynx-js/lynxtron`, along with the `LynxBridge` interface and the `LynxBridgeHandler` type alias. Update the API index, `_meta.json`, and the Communication-Between-Node-And-Lynx guide to link to them.

Change-Id: If7780ac6aa9fbdb2e7b2ecabe47f9d50ae365bb8

- Document the `lynxBridge` singleton and `LynxBridge` API.
- Document the `LynxBridgeHandler` type alias.
- Update English and Chinese API indexes and navigation metadata.
- Explain `lynxBridge` usage and handler replies in the communication guide.
- Update `@lynxtron-examples` showcase dependencies.


lidadating (approved):

Huxpro (changes requested):
Two blocking issues:

1. The callback-style `event.sendReply(...)` example is incompatible with the documented `lynxBridge.handle` implementation. `handle` wraps every handler and immediately calls `event.sendReply(await Promise.resolve(handler(...)))`. A handler that starts callback-based work and returns `undefined` therefore sends an automatic `undefined` reply before the callback later calls `sendReply`, producing a second/ineffective reply. Please make the example return a Promise/result instead, or change and document the runtime contract before recommending explicit `sendReply` from a `handle` callback.

2. These are newly introduced public APIs, but the reference pages and guide provide no first-supported `@lynx-js/lynxtron` version or release boundary. Please add source-backed version metadata (and any relevant runtime/platform compatibility scope) so readers can tell which Lynxtron release exports `lynxBridge`, `LynxBridge`, and `LynxBridgeHandler`.

Huxpro (changes requested):
The updated Promise/return-value guidance correctly fixes the duplicate `sendReply` problem, and CI plus the deploy preview are green on this head.

One blocking documentation boundary remains: the new public `lynxBridge`, `LynxBridge`, and `LynxBridgeHandler` pages still provide no first-supported `@lynx-js/lynxtron` version or release boundary. The implementation landed upstream in `lynx-family/lynxtron@5b3957a` on July 20, 2026, but this PR does not tell users which published package first contains it. Please add source-backed package-version metadata (and any runtime/platform scope if applicable) in both locales before merging.

Huxpro (approved):

PR #1254, opened 2026-08-03
docs: add Autolink Gradle plugin versions

jianliang00:
## Summary

- Pin the Android Autolink settings and build plugins to the published `4.0.1` release.
- Keep the English and Chinese setup examples aligned.

## Validation

- `pnpm exec prettier --check docs/en/guide/autolink.mdx docs/zh/guide/autolink.mdx`
- `pnpm exec cspell lint -c cspell.json --no-must-find-files docs/en/guide/autolink.mdx docs/zh/guide/autolink.mdx`
- `git diff --check`
- `pnpm build`

- Pin Android Autolink settings and build plugins to `4.0.1` in English and Chinese documentation.
- Validate formatting, spelling, whitespace, and site builds.

### DIFF
--- a/docs/en/guide/autolink.mdx
+++ b/docs/en/guide/autolink.mdx
@@ -90,7 +90,7 @@ Enable the settings plugin in `settings.gradle` so library Android projects can 
 
 ```groovy
 plugins {
-  id 'org.lynxsdk.lynx.library-settings'
+  id 'org.lynxsdk.lynx.library-settings' version '4.0.1'
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Huxpro (changes requested):
Blocking on the version boundary and cross-site alignment.

The exact plugin IDs and `plugins { id ... version ... }` integration shown here are not a generic Lynx 4.0 capability. In upstream 4.0.0, the Gradle module explicitly disabled plugin-marker publication and used different IDs; 4.0.1 switches to the `org.lynxsdk.lynx.*` IDs and marker-capable publication. Please state explicitly that this Android Autolink setup requires Lynx 4.0.1 or later.

Please also reconcile that tooling floor with the global site metadata: `docs/public/version.json` still advertises Lynx 4.0.0 as the current/latest 4.0 release. Either update the global patch metadata and dependent references together, or clearly identify 4.0.1 here as a tooling-specific minimum so this page does not silently drift from the rest of the site.

Finally, both deploy previews are currently failing. Please get the Netlify and Cloudflare builds green before this leaves draft or is merged.

PR #1263, opened 2026-08-04
chore(deps): bump @lynx-js/go-web to ^0.8.1

zsy-jason:
Change-Id: Iab0cb1aeced2af84ed2959557bd85c4287c29cd7

- Update the `@lynx-js/go-web` dependency to `^0.8.1`.


Huxpro (approved):
Verified the 0.8.1 release is a focused patch that fixes the Windows deep-link fallback probe without changing the package's public API or compatibility requirements. The dependency and lockfile move together, and CI, CLA, CodeRabbit, and the deploy preview are green.

PR #1268, opened 2026-08-04
fix(version): derive site base from version config

SoonIter:
## Summary

This is the main-branch counterpart of #1221. It derives the Rspress and canonical/OG base from the current entry's `docs_link` in `version.json`, so the `next` docs build keeps using `/next` while sharing the same base source of truth as release builds. It also aligns client-side blog/version links with `withBase()` and updates previous release docs links to their versioned paths.

Note: `netlify.toml` keeps main's existing `/next/* -> /:splat` rewrite; release-only redirect changes from #1221 are intentionally not applied on main.

## Related Links

- https://github.com/lynx-family/lynx-website/pull/1221
- https://github.com/lynx-family/lynx-website/pull/1221#pullrequestreview-4853170441

### DIFF
--- a/docs/public/version.json
+++ b/docs/public/version.json
@@ -20,14 +20,14 @@
       "type": "previous",
       "version_number": "3.9",
       "repo_branch": "release/3.9",
-      "docs_link": "/",
+      "docs_link": "/3.9",
       "release_number": "3.9.0"
     },
     {
       "type": "previous",
       "version_number": "3.8",
       "repo_branch": "release/3.8",
-      "docs_link": "/",
+      "docs_link": "/3.8",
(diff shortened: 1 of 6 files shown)


Huxpro (approved):
The main-branch companion is internally consistent and resolves the cross-branch version-metadata drift: 3.9/3.8 now point to their versioned paths, Rspress and canonical/OG URLs share the same `docs_link`-derived base, and client-side blog links use Rspress `withBase()` rather than a hard-coded `/next`. Main-specific Netlify routing is correctly left unchanged. Build, CLA, CodeRabbit, and deploy-preview checks are green.

Issue #1251, opened 2026-08-02
Follow up unresolved compatibility and API metadata from Lynx 3.9 blog


Huxpro:
PR #1165 was merged with several review blockers still present in the published bilingual Lynx 3.9 documentation. Please address these in a focused follow-up:

- **`flex` shorthand:** `docs/{en,zh}/blog/lynx-3-9.mdx` and `docs/{en,zh}/api/css/properties/flex.mdx` describe omitted `flex-basis` changing unconditionally in Lynx 3.9. The canonical Lynx config defines `enableFlexBasisZeroPercent` as a page switch, supported since LynxSDK 3.8 on Android/iOS/HarmonyOS and defaulting to `false`. Document the switch, default, and actual compatibility/tooling/runtime gate instead of using only a blanket 3.9 badge.
- **`enableNewSticky`:** replace the blanket 3.9 badge with verified per-platform/version compatibility metadata, and make the legacy direct-child restriction consistently conditional on the switch being disabled.
- **Dual-thread Minimizer:** add first-supported package versions to the linked canonical API pages and the blog: `@lynx-js/rspeedy` 0.14.0 for `Minify.mainThreadOptions` / `backgroundOptions`, and `@lynx-js/react-rsbuild-plugin` 0.14.0 for `optimizeBundleSize`.
- **Portal cross-site alignment:** the blog currently names lynx-ui's component as `Overlay` and links `/ui/components/overlay`; the canonical component is `OverlayView` at `/lynx-ui/components/overlay`. Fix both locales while preserving the separate raw `<overlay>` XElement reference.
- **HarmonyOS benchmark claim:** the published “roughly 10%” `TotalTime` / `CPUTime` improvement still lacks a public workload, device/runtime/build configuration, sample size, variance, and measurement method. Add reproducible evidence or remove/soften the numerical result.

Related: #1165 and its blocking review threads.

Huxpro:
Post-merge update: #1260 was merged at `7b93df3`, but it does not resolve the tooling/runtime boundary tracked here. It now states standard sticky behavior and `flex-basis: 0%` as unconditional 3.9 behavior and removes the page-config names entirely.

The source boundaries remain distinct:

- Lynx’s canonical config declares `enableNewSticky` with `defaultValue: false` and `since: 3.8`.
[...]

PR #1269, opened 2026-08-04
docs(api): add config-rsbuild-plugin & Lynx config field reference

upupming:
## Summary

The `@lynx-js/config-rsbuild-plugin` package is published on npm but has no API reference on the website, and its `Config` interface would render empty because the fields live in `@lynx-js/type-config`.

With lynx-family/lynx-stack#3365, the plugin's api-extractor run also produces a `type-config` doc model, so api-documenter renders every public `Config` / `CompilerOptions` field with its TSDoc (description, platform, since, default value), cross-linked from the plugin's `Config` page.

This PR (now additive-only — the unrelated drift refresh of existing pages has been dropped per review):

- registers `packages/rspeedy/plugin-config` in `scripts/update-api-docs.sh` (pkg dirs, build filters, path check)
- adds the generated `config-rsbuild-plugin.*` / `type-config.*` pages for `docs/{en,zh}/api/rspeedy`
- adds `@lynx-js/config-rsbuild-plugin` and `@lynx-js/type-config` entries to `docs/{en,zh}/api/_meta.json`

Pages regenerated from lynx-stack `fe63f6cef` (main), which includes lynx-family/lynx-stack#3365 and its review follow-up lynx-family/lynx-stack#3385.

## Pending before merge

- [x] lynx-family/lynx-stack#3365 merges; the committed pages are regenerated from main for reproducibility
- [x] The invalid `enableCheckExposureOptimize` example and the hard-coded `lynxjs.org/next` link were fixed upstream in lynx-family/lynx-stack#3385; no `lynxjs.org/next` links remain in the generated pages

- Register `@lynx-js/config-rsbuild-plugin` in API documentation generation.
- Publish English and Chinese API reference pages for `@lynx-js/config-rsbuild-plugin` and `@lynx-js/type-config`.
- Document `Config`, `CompilerOptions`, plugin options, and configuration keys with TSDoc details.
- Cross-link plugin configuration fields to `@lynx-js/type-config`.
- Add both packages to the API sidebar metadata.


Huxpro (changes requested):
This is not yet safe to merge as a low-risk API-doc addition.

1. **The committed output is not reproducible from either repository's merged `main`.** This PR says it was generated with lynx-family/lynx-stack#3365, but that upstream PR is still open. `scripts/update-api-docs.sh` regenerates from lynx-stack and deletes local Rspeedy Markdown files that are absent from the generated directory, so running it before #3365 lands would remove the new `type-config` field pages or regenerate the plugin reference differently. Please merge the upstream generator change first, then regenerate this PR from the exact merged upstream SHA.

2. **The regeneration pulls in substantial unrelated API drift without version metadata.** The 155-file diff includes a new `experimental_transformBuiltinAttributeNames` surface for `@lynx-js/react-rsbuild-plugin`; upstream's changelog identifies its first package version as 0.18.2, while the generated API page gives no package-version floor. Similar existing-page drift needs to be split out or fully audited and versioned rather than bundled into this focused addition.

3. **The new reference duplicates the existing global Lynx Config reference without a synchronization/source-of-truth plan.** The generated `Config` page explicitly points users to the existing config reference while adding another page for every field. Compatibility/tooling details around fields such as `enableFlexBasisZeroPercent` are already tracked as unresolved in lynx-website#1251. Please align the canonical reference, related blog/CSS documentation, and this generated field reference before publishing the duplicate surface. Also replace the hard-coded `https://lynxjs.org/next/...` link with a version-aware relative/current-site link so stable-version API pages do not send readers to development docs.

Please regenerate after the upstream dependency merges and keep this PR focused on the new packages, or separately justify and version every unrelated API-doc refresh.

Huxpro (changes requested):
I re-reviewed the regenerated head after lynx-family/lynx-stack#3365 merged. That upstream reproducibility prerequisite is now resolved, but this is still not safe to merge as a low-risk API-reference addition:

1. **Package-version metadata is missing.** The generated pages describe `@lynx-js/config-rsbuild-plugin` 0.2.1 backed by `@lynx-js/type-config` 4.1.3, but expose only runtime `Since: LynxSDK ...` annotations. Those are not the same compatibility boundary as declaration/package availability: for example, `alignMouseEventWithW3C` was added to `@lynx-js/type-config` in 4.1.3 and `enableEventHandleRefactor` in 4.1.2. Please add a visible generated-package snapshot/version and, where necessary, typings/package “since” metadata rather than implying that the SDK floor alone is sufficient.

2. **The primary usage example is not valid against the documented public type.** Both generated `pluginLynxConfig` pages use `enableCheckExposureOptimize`, while `@lynx-js/type-config` 4.1.3 exports an empty `CompilerOptions` and a 33-key `Config` that does not contain this property; the plugin's own type test asserts that shape. Replace the example with a real exported key, or fix the upstream public types and documentation before publishing it.

3. **Unrelated React-plugin API drift remains incomplete and unversioned.** This focused PR still adds the leaf pages for `experimental_transformBuiltinAttributeNames` and `TransformBuiltinAttributeNamesOptions`, but does not update the parent `PluginReactLynxOptions` page or package interface index, leaving the new pages orphaned from the generated API hierarchy. The option also first shipped in `@lynx-js/react-rsbuild-plugin` 0.18.2, which is not stated. Please remove this unrelated drift or regenerate the complete parent/index surface and add the package-version floor in a separately audited change.

4. **The duplicate Lynx Config reference is not yet aligned with the global site.** The generated plugin pages still hard-code `https://lynxjs.org/next/...`, which is wrong for stable-version builds, and the new field pages create a second full source of truth while known global blog/API inconsistencies remain open in #1251 (for example, `enableFlexBasisZeroPercent` is correctly conditional/default-false here but is still described unconditionally elsewhere). Please use a version-aware current-site link and land the related cross-site corrections together, or define a single canonical/synchronized reference before publishing this duplicate surface.

CI and the deploy preview are green, but they do not resolve these documentation, compatibility, and cross-site consistency blockers.

Huxpro (changes requested):
The unrelated React-plugin drift has been removed, and the generated output is now additive-only, but three substantive documentation/compatibility blockers remain on this head:

1. **Package-version metadata is still missing.** These pages publish a snapshot of `@lynx-js/config-rsbuild-plugin` and `@lynx-js/type-config`, but expose only runtime `Since: LynxSDK ...` annotations. Runtime SDK availability and declaration/package availability are different boundaries. Please make the generated package snapshot/version visible and add typings/package “since” metadata where a field was introduced after the runtime capability.

2. **The primary example is invalid against the documented public type.** Both generated examples pass `enableCheckExposureOptimize`, while the generated `Config`/`CompilerOptions` reference does not export that key. Replace it with a real public configuration property, or correct the upstream public types before publishing the example.

3. **The duplicate configuration reference is not aligned with the global site.** The generated `Config` pages still hard-code `https://lynxjs.org/next/api/lynx-config/config-reference.html`, which is wrong when these docs are built for a stable version. This PR also creates a second full reference for the same Lynx Config fields without establishing a canonical/synchronized source, while related descriptions remain duplicated elsewhere on the website. Please use a version-aware current-site link and land the cross-site alignment together, or define one source of truth before publishing this duplicate surface.

CI and the deploy preview are green, but they do not resolve these versioning, example-validity, and cross-site consistency issues.

Huxpro (approved):
Acceptable

upupming:
Resolving the remaining CodeRabbit threads — all six point at content this repo does not own.

`docs/{en,zh}/api/rspeedy/type-config.*` and `config-rsbuild-plugin.*` are api-documenter output generated from the `@lynx-js/type-config` TSDoc in lynx-stack. Editing them here would be overwritten by the next `scripts/update-api-docs.sh` run, and CodeRabbit's own path instruction says to fix the source and regenerate.

Tracking upstream:

- `a11y-id based` → `a11y-id-based` (`enableA11y`)
- `enableTextOverflow` describes Harmony text rendering but lists only Android/iOS as supported platforms — needs the platform metadata reconciled
- `cancelled` vs `canceled` inconsistent between `enableEndGestureAtLastFingerUp` and neighbouring fields
- zh pages carry English TSDoc; localized API text has to come from the upstream source or a locale generation input

These will land here automatically the next time the pages are regenerated.

PR #1270, opened 2026-08-04
fix(blog): read the blog from the developing version, not this build's copy

Huxpro:
## Problem

Two symptoms on the live 4.0 site, one cause:

- The homepage badge announces **Lynx × Rspack 2.0** (2026-06-30). It has never announced 3.9.
- `/4.0/blog/lynx-3-9` is a **404**, and every card on `/4.0/blog/` links to `/4.0/blog/...`.

This branch was cut before the 3.9 post landed, and blog posts are never backported — so this build's own copy of the blog stops at the cut. Anything that reads the blog out of this build reads a truncated list.

## Changes

**Blog links point at the developing version again.** They used to, for exactly this reason; that was replaced with a per-version base, which re-versions the blog and sends readers to pages this build doesn't have. The base is now derived from `version.json`'s `developing` entry rather than hardcoded `/next`, so it follows a rename.

**Cross-version blog hrefs no longer go through `BaseLink`.** With this build's base being `/4.0`, `BaseLink` re-applies it (`/4.0/next/blog/...`) and then intercepts the click into a route this app doesn't have. A plain anchor does the page load that actually gets there. Verified both failure modes in a browser before fixing.

**The hero badge reads the canonical blog.** It took the newest post from `usePages()` — the same truncated page data. It now reads the first entry of the developing build's RSS feed, which is same-origin from every version.

The feed carries `title` rather than `badge_text`, so the badge shows the full post title. Deliberate trade: naming the right release matters more than the shorter wording. The local result stays the fallback while the feed is in flight and if it can't be read, so the badge always renders. A pinned `filename` or an `externalLink` is left alone.

## Verification

The mechanism was verified end-to-end on `main` against a build configured to look like a release build (base `/3.9`, blog cross-version), driving a real browser with the **real** `/next/` feeds served through request interception:

| scenario | badge text | click target |
|---|---|---|
| en, feed ok | `Lynx 3.9: Autolink, Web-Aligned CSS, and ReactLynx Portal` | `/next/blog/lynx-3-9` |
| zh, feed ok | `Lynx 3.9：Autolink、更贴近 Web 的 CSS、ReactLynx Portal` | `/next/zh/blog/lynx-3-9` |
| feed returns 500 | local fallback | `/next/blog/lynx-3-9` |
| feed returns malformed XML | local fallback | `/next/blog/lynx-3-9` |

Blog index cards on that build resolve to `/next/blog/...` with no doubled base, and clicking the badge does a full page load rather than an SPA navigation.

On this branch: `tsc --noEmit` introduces no new errors over the unmodified branch. A full production build was run on the `main` counterpart, not here — CI covers this one.

Not touched: the `rel="alternate"` language link still points within the current version, which matches today's behaviour and is what hreflang wants.

## Related

- #1266 — the `main` counterpart.
- #1221 — overlaps here. That PR's base fix is right and needed; this only asks that the blog keep a base of its own rather than following `withBase()`. Happy to fold this into it instead if that's easier to land.

---
_Generated by [Claude Code](https://claude.ai/code/session_017UqU8yvoJn5uGFj9wF6KeC)_

### DIFF
--- a/shared-route-config.ts
+++ b/shared-route-config.ts
@@ -1,6 +1,37 @@
 /**
  * Sub-sites and shared docs configuration
  */
+import versionJson from './docs/public/version.json';
+
+/** This build's Rspress base as a URL prefix — matches rspress.config.ts. */
+export const SITE_BASE_PREFIX = `/${versionJson.current_version}`;
+
+const developingDocsLink = versionJson.versions.find(
+  (version) => version.type === 'developing',
+)?.docs_link;
+
+if (!developingDocsLink) {
+  throw new Error('Missing docs_link for the developing version');
+}
+
+/**
+ * Base prefix for blog links, on every version of the site.
+ *
+ * The blog is not versioned content: this branch was cut before its own
+ * release post landed, and posts are never backported, so this build's copy
+ * of the blog is missing everything published since the cut. Blog links
+ * therefore point at the developing version, which is the single source of
+ * truth. Derived from `version.json` rather than hardcoded so it follows the
+ * developing entry if `next` is ever renamed.
+ */
+export const BLOG_BASE =
+  developingDocsLink === '/' ? '' : developingDocsLink.replace(/\/$/, '');
+
+/**
+ * Whether this build's blog links leave its own base. True on every release
+ * build; false on the developing build, where in-app routing just works.
+ */
+export const BLOG_IS_CROSS_VERSION = BLOG_BASE !== SITE_BASE_PREFIX;
 
 /**
  * Metadata for each subsites. This is used to
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -4,4 +4,5 @@ export {
   type LatestBlogConfig,
   type LatestBlogResult,
 } from './use-latest-blog';
+export { useCanonicalLatestBlog } from './use-canonical-latest-blog';
 export { useTiltEffect } from './use-tilt-effect';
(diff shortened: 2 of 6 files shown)


Huxpro:
This provides the missing `release/4.0` counterpart, but I would keep it blocked/draft until the rollout is made coherent across the related PRs and sites:

- **Stack or fold this into #1221.** Both PRs modify `shared-route-config.ts`, `src/lib/utils.ts`, and `theme/index.tsx`. This branch currently defines `SITE_BASE_PREFIX` as `/${versionJson.current_version}` (`/4.0`), while #1221 changes the actual 4.0 Rspress base to the root (`/`). Although the cross-version boolean is true in either case, landing the PRs independently is order-dependent and leaves duplicated base logic with conflicting definitions. The final release/4.0 diff should derive the site base and blog base from the same shared version metadata.

- **The 3.9 release remains affected.** #1266 identifies the stale hero badge on `/3.9/`, but `main` + this 4.0 companion do not update `release/3.9`. Please add the corresponding backport or explicitly coordinate it before treating the cross-version fix as complete.

- **The RSS source is still inconsistent.** The new hook consumes `/next/rss/...` as canonical, but the feed generator still emits unversioned item links that redirect to missing pages. Please align the RSS `siteUrl`/item URLs with the canonical blog base rather than correcting the path only inside this one consumer.

CI is still running and the PR is a draft, so no merge was attempted. I cannot submit a formal request-changes review because this PR is authored by the authenticated account.

PR #1274, opened 2026-08-04
docs(desktop): element backgrounds occlude native views; detaching is an alternative to cover-view

Huxpro:
Two gaps found while embedding a native editor (Scintilla) in a Lynxtron desktop app on Lynx 4.1 / Clay macOS. Each cost real debugging time, and neither is derivable from what the pages say today.

## 1. Element backgrounds hide the platform control

`custom-component-desktop` already says `z-index` cannot reliably interleave Lynx content with a Native UI control. That reads as *"you cannot put Lynx content above it"*. The sharper problem is the reverse — **a `background-color` on any element whose box overlaps the control hides the control entirely**, because Clay promotes the background to its own layer and the platform control composites below all of them.

That includes the **application root**, which is an ordinary element rather than the render tree's root — and which is exactly where an app is tempted to put its ground colour.

What makes this expensive is how it presents:

- attachment succeeds
- `OnLayoutChanged` reports the correct rect
- content is in the document and readable back
- the region is blank

Every signal says the control is healthy, so it reads as a compositor bug. In our case it survived a `git bisect` pointing at an unrelated commit, and I was about to file an engine issue against it.

Worse, putting the ground on the app root *looks* like a fix: it renders on some launches and not others, because it is a layer-order race. Only moving the background to the **host window** — the one surface below both Lynx content and the control — is stable.

The added note states the rule and where the background belongs.

## 2. Detaching is an alternative to `<cover-view>`

`cover-view` already recommends a plain `<view>` with `position: fixed` when the overlay only needs to cover Clay-rendered content. It stops one step short for a surface that **replaces** a native view rather than floating over it: detach the platform control while that surface is up, and there is nothing left to cover — so a regular `<view>` is enough and no platform overlay is needed.

This is not a micro-optimization. A full-page surface that takes over from an embedded editor is a common shape, and routing it through a platform overlay costs behaviour it does not need. The added sentence is what points a reader at the cheaper path.

## Notes

- Both changes are in `en` and `zh`.
- Documentation only; no code or compat-data changes.
- Related behaviours we also measured are *not* in this PR because they look like defects rather than contracts, and documenting them would risk freezing them — they are filed separately against `lynx-family/lynxtron`.

- Clarify how element and application-root backgrounds can occlude native controls, and recommend using the host window background.
- Explain when to detach native controls and use `<view>` instead of `<cover-view>`.

### DIFF
--- a/docs/en/api/elements/built-in/cover-view.mdx
+++ b/docs/en/api/elements/built-in/cover-view.mdx
@@ -13,6 +13,8 @@ import { APISummary, APITable, ClayMacOSOnly, ClayWindowsOnly } from '@lynx';
 
 For overlays that only need to cover other Clay-rendered content, prefer a regular `<view>` with `position: fixed`.
 
+This includes surfaces that *replace* a native view rather than float over it. Detach the platform control while the surface is up and there is nothing left to cover, so a regular `<view>` is enough and no platform overlay is needed.
+
 ## Usage
 
 Position and size `<cover-view>` with the same CSS layout properties used by `<view>`. Its children are rendered into the platform overlay surface within those bounds.
--- a/docs/en/guide/custom-native-component/custom-component-desktop.mdx
+++ b/docs/en/guide/custom-native-component/custom-component-desktop.mdx
@@ -132,6 +132,10 @@ The Native UI rendering backend attaches a platform control and updates its layo
 
 :::tip Covering native views or child windows
 Platform views or child windows attached by the Native UI backend are placed above Clay's main rendering surface, so regular Lynx `z-index` cannot reliably cover them. Keep the platform control inside the renderer host returned for its `lynx_view_t`, and place overlapping Lynx content inside a [`<cover-view>`](/api/elements/built-in/cover-view). If the UI requires arbitrary clipping, transforms, or interleaving with regular Lynx content, use the Texture backend instead.
+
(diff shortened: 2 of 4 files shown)


Huxpro:
Blocking compatibility concern before merge:

The detaching guidance is useful, but the background-occlusion statement is currently written as a shared Clay/Native UI rule even though the PR description says it was measured only on **Lynx 4.1 / Clay macOS**.

This page covers both Clay macOS and Clay Windows. The existing Windows model is a child `HWND` attached to the renderer host, and the surrounding docs say native child windows sit above Clay's main rendering surface. The new paragraph then generalizes in the opposite direction—**any** overlapping element background hides the control and the control is below **all** such layers—without independent Windows verification or a platform/version qualifier.

Please either:

1. verify and document the behavior on both Clay macOS and Clay Windows, including the applicable Lynx version floor; or
2. scope the warning explicitly to **Clay macOS / Lynx 4.1** (ideally in the macOS-specific section or with the corresponding platform/version badge), and avoid presenting the host-window workaround as the only cross-platform rule.

Keep the English and Chinese copies aligned after the scoping change. The two unresolved Copilot wording comments should also be addressed. I cannot submit `REQUEST_CHANGES` because this account is the PR author, so I am leaving the blocker here instead.

PR #1276, opened 2026-08-06
docs(react): remove the build-time macros page, rehome the useful switches

upupming:
The page mostly documented implementation details. Removed it and moved the useful switches to where readers would look for them.

- Public macros: `Document.built-in-macros` now covers all eight globals declared by `@lynx-js/react` (it previously documented only `__BACKGROUND__` and `__MAIN_THREAD__`). The old URL redirects there.
- `REACT_ALOG` / `REACT_ALOG_ELEMENT_API`: new Diagnostic Logs page.
- `REACT_DEVTOOL`: Preact DevTools panel page.
- Profiling instrumentation is a build-time switch: noted on the Profiling page, which now links to `rspeedy/runtime-profiling` for `performance.profile`.

The rehomed switches carry package-version notes where support has a floor.

- Remove the “Build-time Macros” page and redirect its URLs to the built-in macros reference.
- Document all eight `@lynx-js/react` globals on the built-in macros reference.
- Document `REACT_ALOG` and `REACT_ALOG_ELEMENT_API` on the Diagnostic Logs page.
- Document `REACT_DEVTOOL` on the Preact DevTools panel page with required plugin versions.
- Document profiling instrumentation defaults and production runtime profiling guidance.


Huxpro (changes requested):
This removal needs a coordinated public-API decision before merge.

The PR says these macros are implementation details and “not exposed,” but the current published ReactLynx types still declare several of them as application-visible globals (`__ENABLE_SSR__`, `__FIRST_SCREEN_SYNC_TIMING__`, `__GLOBAL_PROPS_MODE__`, and `__DISABLE_CREATE_SELECTOR_QUERY_INCOMPATIBLE_WARNING__`), and `ReactWebpackPlugin` injects values for all six through `DefinePlugin`. That makes a silent documentation deletion inconsistent with the upstream type surface and build behavior.

Please either:

1. coordinate the upstream type/API change that makes these names genuinely internal, including package/version or migration implications, and link it here; or
2. retain an explicit section stating that these globals are unsupported/internal and may change, rather than removing previously published behavior without replacement.

The English and Chinese pages are aligned, and CI is green, but this is a non-additive public-documentation removal with unresolved cross-repository consistency.

colinaaa (approved):

Huxpro (changes requested):
Thanks for rehoming the actionable switches; that addresses the discoverability part of the earlier review. I still can’t treat this head as merge-safe for two reasons.

1. The cross-repository API/docs mismatch remains. `@lynx-js/react` still publishes ambient declarations for `__DEV__`, `__PROFILE__`, `__REACT_DEVTOOL__`, `__ALOG__`, `__ALOG_ELEMENT_API__`, `__ENABLE_SSR__`, `__FIRST_SCREEN_SYNC_TIMING__`, `__GLOBAL_PROPS_MODE__`, and other compile-time globals, and `ReactWebpackPlugin` still defines these values. The new redirect target (`Document.built-in-macros`) documents only `__BACKGROUND__` and `__MAIN_THREAD__`. Please either coordinate the upstream type/API decision that makes the removed names genuinely internal, or retain an explicit unsupported/internal reference so users are not silently redirected from a broader published surface to a page that omits it.

2. The newly rehomed build switches are version-sensitive, but the new pages give no toolchain/package floor. Upstream has known introduction points: production `REACT_DEVTOOL` support landed in `@lynx-js/react-webpack-plugin@0.9.5`, while `REACT_ALOG_ELEMENT_API` became a separate opt-in in `@lynx-js/react-webpack-plugin@0.7.4` (released via the corresponding React/Rspeedy stack). Please add concise package-version/provenance notes in both EN and ZH rather than presenting these instructions as universally available.

CI and the EN/ZH edits are aligned, but this remains a non-additive documentation removal with unresolved public-type/docs consistency and missing version scope.

upupming:
@Huxpro 

1. I have double checked the macros, only __DEV__, __PROFILE__, __REACT_DEVTOOL__ are publicly exposed in https://github.com/lynx-family/lynx-stack/blob/main/packages/react/types/react.docs.d.ts, all others are exposed only in `packages/react/runtime/types/types.d.ts` which are private only. So removing these macros is safe.

Huxpro (changes requested):
The new version notes address the version-scope blocker from my previous review: the EN/ZH DevTools and diagnostic-log pages now carry explicit `@lynx-js/react-rsbuild-plugin` / `@lynx-js/react-webpack-plugin` floors, and CI is green on this head.

One substantive blocker still remains, so I’m not comfortable merging the removal yet. The redirected `Document.built-in-macros` reference documents only `__BACKGROUND__` and `__MAIN_THREAD__`, while the current `@lynx-js/react` published global types still declare `__DEV__`, `__PROFILE__`, `__REACT_DEVTOOL__`, `__ALOG__`, `__ALOG_ELEMENT_API__`, `__ENABLE_SSR__`, `__FIRST_SCREEN_SYNC_TIMING__`, `__GLOBAL_PROPS_MODE__`, `__DISABLE_CREATE_SELECTOR_QUERY_INCOMPATIBLE_WARNING__`, etc.; `ReactWebpackPlugin` still injects these values via `DefinePlugin` as well.

Please resolve that public-type/docs mismatch before deleting the broader page—either by coordinating the upstream type/API change that makes these names genuinely internal, or by retaining a concise unsupported/internal reference for the still-published globals. The latest commit fixes the version metadata, but not this cross-repository consistency issue.

upupming:
@Huxpro No, all these __ALOG__, __ALOG_ELEMENT_API__, __ENABLE_SSR__, __FIRST_SCREEN_SYNC_TIMING__, __GLOBAL_PROPS_MODE__, __DISABLE_CREATE_SELECTOR_QUERY_INCOMPATIBLE_WARNING__ are not publicly exposed by reactlynx. But for __DEV__, __PROFILE__, __REACT_DEVTOOL__ I think you are right.

Huxpro (changes requested):
The latest commit resolves the public-surface mismatch from my previous review: the Built-in Macros reference now matches the eight globals intentionally exposed in `packages/react/types/react.docs.d.ts` (`__JS__`, `__BACKGROUND__`, `__LEPUS__`, `__MAIN_THREAD__`, `__DEV__`, `__PROFILE__`, `__REACT_DEVTOOL__`, and `__LAZY_BUNDLE_FETCHER__`), and the EN/ZH changes stay aligned.

One compatibility/version blocker remains on the newly documented public macro: `__LAZY_BUNDLE_FETCHER__` is version-sensitive, but its new API-reference entry currently has no version or engine scope. This macro arrived with the FetchBundle lazy-loader stack released in `@lynx-js/react@0.123.0` / `@lynx-js/react-rsbuild-plugin@0.18.0`; in that stack, `FetchBundle` is selected when `engineVersion >= 3.9`, otherwise the legacy `QueryComponent` path is used. Please record that package floor and engine gating/provenance in both locales so users do not read this as an unversioned, universally available macro.

Also, `Document.built-in-macros.mdx` is explicitly marked generated (`Do not edit this file directly`). Please make this addition through the authoritative typedoc source / regeneration path, or otherwise ensure the generated artifact remains reproducible rather than becoming a website-only fork of the source docs.

Given that this PR still removes an existing user-facing page, I’m not comfortable merging until that compatibility scope and generated-source alignment are explicit.

Huxpro (changes requested):
The new version note fixes the missing package/engine floor, but this head still has two substantive consistency/correctness blockers:

1. The `__LAZY_BUNDLE_FETCHER__` description is incomplete about how the value is selected. Current `resolveLazyBundleFetcher()` checks `REACT_LAZY_BUNDLE_FETCHER` before falling back to the engine-version-derived value: `REACT_LAZY_BUNDLE_FETCHER=QueryComponent` can force the legacy loader even when `engineVersion >= 3.9`; `FetchBundle` may be forced only when the declared engine version is >= 3.9 (otherwise the plugin throws). Please document that override instead of saying the value is simply `FetchBundle` for >=3.9 and `QueryComponent` otherwise.

2. The generated-source mismatch from the previous review is still unresolved. Both `Document.built-in-macros.mdx` files say they are generated by `@lynx-js/tool-typedoc` and must not be edited directly, while the authoritative `lynx-stack/packages/react/types/react.docs.d.ts` currently contains the macro declaration/description but not this new version-scope text. Regenerating the API reference can therefore erase the fix. Please land the documentation in the typedoc source and regenerate (or move the website-owned compatibility detail to a non-generated canonical page and link it from the generated reference).

CI is green and EN/ZH remain aligned, but because this PR removes an existing user-facing page and redirects it to this generated reference, I don't think a website-only generated-file patch is merge-safe yet.

Huxpro (approved):
The latest two commits resolve the substantive blockers from my previous reviews.

- The lazy-bundle compatibility contract is now documented on the non-generated Code Splitting page: the default `engineVersion` selection, the `REACT_LAZY_BUNDLE_FETCHER` override in both directions, the `engineVersion >= 3.9` guard for `FetchBundle`, and the `@lynx-js/react@0.123.0` / `@lynx-js/react-rsbuild-plugin@0.18.0` introduction floor all match the current implementation and changelog.
- The generated Built-in Macros reference now points to that canonical compatibility section instead of trying to carry the version/selection rules itself. Also, the repository's own `scripts/update-api-docs.sh` explicitly excludes `docs/{en,zh}/api/react` from the automated regeneration pipeline because that reference is bespoke, so the earlier concern about this being overwritten by the normal API-doc sync does not apply here.
- The eight documented globals match the current `@lynx-js/react` public docs declarations, the DevTools/ALog package floors are backed by their release history, and EN/ZH remain aligned.
- GitHub Actions CI and the Netlify preview are green on this exact head. (The optional Cloudflare preview is also failing on recently merged PRs such as #1279, so I am not treating that as PR-specific.)

Approving this head. I am intentionally not auto-merging it: this PR deletes/replaces an existing guide and redirects its URL, so although the review issues are resolved, it is not an additive-only change under the conservative auto-merge policy.

docs/zh/react/diagnostic-logs.mdx
```diff
@@ -0,0 +1,41 @@
+---
+description: '打开 ReactLynx 的详细框架日志，排查真机上的问题。'
+---
+
+# 诊断日志
```
Huxpro: 我并不希望把这一页作为给开发者的指南里的一个顶层的目录。因为它对太多开发，大部分开发者可能是没有用的。但是我们现在目前好像也并没有一个叫 Advance 或者 Internal 的地方，所以我觉得这里需要思考一下有没有什么东西可以把它先藏起来，或者跟它 group 在一起。

Huxpro (changes requested):

Huxpro (approved):
Reviewed the new `89b10ce` follow-up. Folding the ReactLynx ALog guidance into the existing global **Handle Errors** page is a better fit than adding a separate top-level React guide, and it resolves the duplication/navigation concern without losing the documented switches. The EN/ZH changes stay aligned, the existing `REACT_ALOG_ELEMENT_API` package floor is preserved, and CI/preview are green on this exact head. No new substantive blocker from this revision.

PR #1277, opened 2026-08-06
feat: add snapshot

HuJean:
Change-Id: Ie5e0e0e82ccaf93996bc5b6ec82f28de78db60b3

- Document ReactLynx Snapshot compilation, runtime behavior, and author guidance.
- Add Snapshot to English and Chinese React documentation navigation.
- Describe Snapshot as a main feature in English and Chinese introductions.
- Clarify first-screen rendering flow in English and Chinese lifecycle documentation.
- Link Snapshot optimization sections to the Snapshot compilation documentation.


Huxpro (changes requested):
Requesting changes before merge because this adds public architecture guidance with missing version scope and leaves related site content inconsistent:

1. **Version and lazy-bundle compatibility:** The generated-code section is described as the current ReactLynx compiler output and uses the `snapshotCreatorMap` / SlotV2-era model, but Snapshot generation and runtime compatibility have changed across `@lynx-js/react` versions. In particular, lazy snapshot registration introduced a one-way lazy-bundle compatibility boundary: a new runtime can load older bundles, while an older runtime cannot load newly produced lazy bundles. Please state which `@lynx-js/react` version or compiler generation this page describes and document the compiler/runtime pairing requirement for lazy bundles (using the standard version/compatibility metadata where applicable). The general warning that internal output may change is not enough for readers using this page to inspect or diagnose bundles.

2. **Cross-site alignment:** Both the English and Chinese Lynx 3.9 release posts still say that Snapshots will be covered in a later post. Now that the canonical Snapshot page exists, update both posts to link to it so the global blog and reference documentation stay aligned.

3. **Locale alignment:** This PR clarifies the first-screen rendering and synchronization flow only in `docs/zh/react/lifecycle.mdx`. Apply the same semantic correction to the English lifecycle page; a core architecture explanation should not diverge by locale.

4. **Canonical terminology:** Both introductions still call this optimization “JSX constant folding” earlier on the page while the feature list now calls it “Snapshot.” Please use one canonical term, or explicitly explain their relationship, consistently in English and Chinese.

Please keep all related English and Chinese changes aligned.

Huxpro (changes requested):
Re-reviewing the new head `cd3b01f`. The rewrite makes the Snapshot explanation clearer, and the earlier terminology issue in the introductions is now resolved. I still don't think this is safe to merge yet:

1. **Version / compatibility scope is still missing.** This page documents a concrete Snapshot codegen/runtime model (`snapshotCreatorMap`, `__DynamicPartSlotV2`, generated creator/updater behavior) but still does not state which `@lynx-js/react` / compiler generation it describes, nor the compiler/runtime compatibility requirement for lazy bundles. The generic warning that internals may change is not sufficient for architecture/debugging guidance. Please add an explicit supported/version scope (and the standard version/compatibility metadata where applicable).

2. **The global 3.9 blog remains stale.** Both localized Lynx 3.9 posts still say Snapshots will be covered "in a later post". Now that this is intended to be the canonical Snapshot page, update those EN/ZH posts to link here so duplicated site content does not contradict the new docs.

3. **Locale lifecycle alignment is still incomplete.** The PR still changes only `docs/zh/react/lifecycle.mdx`; the equivalent first-screen/synchronization correction is absent from the English lifecycle page. Please keep the architecture explanation aligned across locales.

4. **The new pseudocode is internally inconsistent.** The Snapshot definition's updater reads `ctx.__elements` / `ctx.__values`, while the later `renderFirstScreen` / `applyAttributePatch` pseudocode writes `instance.elements` / `instance.values` and then calls that updater. As written, those examples do not compose. Please either use the real prefixed fields consistently or make the simplified definition/updater use the same unprefixed fields throughout, in both locales.

The latest commit changes only the EN/ZH Snapshot pages, so items 2 and 3 above were not addressed by this update.

docs/en/react/snapshot.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+:::info
+
+Snapshot is unrelated to snapshot testing or the `getSnapshotBeforeUpdate()` method on React class components. It is an internal ReactLynx implementation detail. Application code should continue to use regular JSX and should not create or manipulate Snapshots directly.
+
+:::
+
+## Why Snapshot
+
+A regular JSX transform keeps the element structure, static values, and dynamic expressions in the runtime output. On every render, `_jsx()` creates virtual nodes (VNodes) that describe the UI structure. The ReactLynx reconciler then compares these nodes level by level to decide which host elements to reuse, create, or update. This process is known as reconciliation.
```
upupming, line 58: What is `_jsx`? How about `React.createElement`/`jsx`/`jsxDev`?

We can refer https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

Huxpro (changes requested):
Current head improves the guide's organization, but the existing blockers remain: missing Snapshot compiler/runtime version scope and lazy-bundle compatibility guidance; stale EN/ZH 3.9 blog text that still says Snapshot will be covered later; a lifecycle correction present only in Chinese; and pseudocode that writes `instance.elements` / `instance.values` while the updater reads `ctx.__elements` / `ctx.__values`. Please resolve these before merge. CI and deploy checks are green; this is blocked on documentation correctness and compatibility.

Huxpro (changes requested):
Re-reviewing `bea233bf`: the previous blockers are substantially addressed — EN/ZH 3.9 now link the canonical Snapshot guide, the lifecycle edit is present in both locales, the pseudocode fields are aligned, and the 0.115/0.120 compatibility boundaries are now stated. Two compatibility/cross-site issues still keep this from the low-risk merge bar:

1. **The lifecycle wording still describes a tree-to-tree comparison that the new Snapshot guide does not.** Both lifecycle pages now say the background thread builds a tree, then “compares the tree structures from both threads and synchronizes their differences.” The Snapshot guide in this same PR instead describes initial background reconciliation creating the logical instance tree, followed by hydration establishing the mapping to the main-thread instances, after which background reconciliation emits updates that the main thread applies. Please make the lifecycle summary use that same model in EN/ZH rather than implying a generic comparison between two independently rendered trees.

2. **The new SlotV2 compatibility note omits the current compatibility escape hatch.** Main already documents `pluginReactLynx({ compat: { legacySlot: true } })`: it intentionally emits the pre-SlotV2 form so a newer compiler can target runtimes `< 0.120.0`, while the runtime supports both forms. The Snapshot page currently says its generated output reflects “0.120.0 or later” and gives a blanket upgrade-to-same-or-newer recommendation without mentioning this supported path. Since this section is specifically about standalone lazy-bundle compatibility, please acknowledge/link `CompatVisitorConfig.legacySlot` and distinguish the 0.115 `snapshotCreatorMap` boundary from the 0.120 SlotV2 boundary.

CI is green on this head; these are documentation correctness/compatibility blockers only.

Huxpro (changes requested):
Current head `e0b88781` still has one compatibility blocker. The new EN/ZH Code Splitting warning says a standalone lazy-bundle consumer should always be upgraded to the same or a newer `@lynx-js/react` version. That conflicts with the site's canonical `CompatVisitorConfig.legacySlot` API reference, which explicitly supports newer compiler output targeting runtimes `< 0.120.0` by emitting the pre-SlotV2 form. Please distinguish the hard 0.115 `snapshotCreatorMap` incompatibility boundary from the configurable 0.120 SlotV2 boundary, and link `legacySlot` from the warning. CI is green on this head; I do not see another blocker once this compatibility wording is aligned.

Huxpro:
Re-reviewing current head `cf82e763`: the compiler-informed-rendering reframing is clearer and the EN/ZH edits stay aligned, but the remaining compatibility blocker from the previous head is unchanged. The Code Splitting warning still says a newly built standalone lazy bundle requires upgrading the Consumer to the same or newer `@lynx-js/react`. That conflicts with the canonical `CompatVisitorConfig.legacySlot` API reference, which explicitly supports newer compiler output targeting runtimes `< 0.120.0` by emitting the pre-SlotV2 form, while the runtime accepts both forms. Please distinguish the hard 0.115 `snapshotCreatorMap` boundary from the configurable 0.120 SlotV2 boundary and link `legacySlot` from the warning. CI and deploy preview are green on this head; I do not see another blocker once this compatibility wording is aligned.

Huxpro (changes requested):
Re-reviewing current head `73c0b26e`. The Compiler-Informed Rendering reframing is coherent, but two documentation correctness/compatibility blockers remain:

1. The simplified Snapshot updater reads `ctx.__elements` / `ctx.__values`, while the later first-screen and update pseudocode writes `instance.elements` / `instance.values` and passes that `instance` directly to `definition.update(...)`. Those examples no longer compose as written. Please use one instance shape consistently in both EN/ZH snippets.

2. The standalone lazy-bundle warning still broadly says the Consumer must be upgraded to the same or newer `@lynx-js/react`. The canonical `CompatVisitorConfig.legacySlot` reference explicitly supports newer compiler output targeting runtimes `< 0.120.0` by emitting the pre-SlotV2 form, including the standalone-lazy-bundle case. Please distinguish the hard 0.115 `snapshotCreatorMap` compatibility boundary from the configurable 0.120 SlotV2 boundary and link `legacySlot` from this warning.

GitHub CI is still running on this head; these two issues are blockers independently of CI.

Huxpro (approved):

Huxpro (approved):
Updated the collapsible IR definition blocks to reuse the learn/gallery details style in both English and Chinese. CI and preview checks pass.

PR #1281, opened 2026-08-07
docs(gallery): let the tutorial's list fill its wrapper

Huxpro:
## What

The gallery tutorial's CSS leaves a 48px black band under the grid — visible in
the `<Go>` previews on this very page.

## Why

`.list` is `height: calc(100% - 48px)` inside a `.gallery-wrapper` that is
`height: 100%` with a black background, and the scrollbar derives its viewport
from `lynx.__globalProps.screenHeight - 48`.

That 48 is **Lynx Explorer's title bar** — which this tutorial's own QR schema
turns on with `?bar_color=000000&back_button_style=dark&title=Popular Furniture`.
But `100%` is already the page area *below* that bar, so the code subtracts it a
second time. And a shorter list would not move content out from under a *top* bar
anyway, so the 48px only ever appears as dead space at the bottom:

- on **Lynx for Web**, including the previews on this page;
- in **Explorer's fullscreen** mode;
- and **under the title bar** itself.

## How

- `.list` fills its wrapper (`height: 100%`).
- The scrollbar takes its viewport from `event.detail.listHeight` — the list's own
  box — so it is correct whatever area the list ends up with. Lynx for Web reports
  `0` there, so the page height stands in; once the list fills the page it is the
  same number.
- The `:::details __globalProps` aside is replaced with one about `listHeight`,
  since the snippet no longer uses `globalProps`.
- Both `en` and `zh` are updated together.

## Verification

Measured through `@lynx-js/web-core` in headless Chromium against the example
these snippets are quoted from, `GalleryComplete`:

| | before | after |
| --- | --- | --- |
| `.gallery-wrapper` height | 700 | 700 |
| `.list` height | 652 | **700** |
| band at bottom | **48px** | **0** |
| scrollbar inline style | `height: 60.1px; top: 51.5px` | `height: 69.3px; top: 53.4px` |

## Related

Depends on lynx-family/lynx-examples#396, which fixes the example this page's
previews load. Merging this alone would leave the prose ahead of the code.

The same defect is inherited by the Vue Lynx and Octane ports:

- Huxpro/vue-lynx#360
- octanejs/octane#616

- Set the gallery `.list` height to `100%` in the English and Chinese tutorials.
- Pass `event.detail.listHeight` to scrollbar sizing, with a Lynx for Web page-height fallback.
- Replace `__globalProps` documentation with `listHeight` behavior details.

### DIFF
--- a/docs/en/learn/gallery.mdx
+++ b/docs/en/learn/gallery.mdx
@@ -100,7 +100,7 @@ Since the focus of this tutorial is not on how to style your UI, you may just sa
   padding-bottom: 20px;
   padding-left: 20px;
   padding-right: 20px;
-  height: calc(100% - 48px);
+  height: 100%;
   list-main-axis-gap: 10px;
   list-cross-axis-gap: 10px;
 }
@@ -340,7 +340,8 @@ Similar to the `bindtap` event used to add the like functionality, we add the [`
 const onScroll = (event: ScrollEvent) => {
   scrollbarRef.current?.adjustScrollbar(
     event.detail.scrollTop,
-    event.detail.scrollHeight
+    event.detail.scrollHeight,
+    event.detail.listHeight || SystemInfo.pixelHeight / SystemInfo.pixelRatio
   );
 };
 ...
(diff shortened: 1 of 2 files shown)


Huxpro (commented):
Blocking on three consistency issues before this is merge-safe:

1. The updated tutorial snippet still types the handler as `ScrollEvent`, but it now reads `event.detail.listHeight`. The companion example PR correctly switches this to `ListScrollEvent` because the generic `ScrollEvent` detail is `BaseScrollInfo` and does not expose `listHeight`. Please update the EN/ZH snippets (and any shown imports if applicable) so the tutorial code type-checks against the documented API.

2. This replacement removes the tutorial's only `__globalProps` explanation/link even though a later gallery step still uses `lynx.__globalProps.screenWidth`. Please keep that later dependency documented rather than making the page less self-contained. Also, the fallback here is `SystemInfo.pixelHeight / SystemInfo.pixelRatio` (screen height in CSS px), so calling it “page height” should be qualified.

3. As noted in the PR body, this documentation quotes behavior from `lynx-family/lynx-examples#396`, which is still open. Merging this first would put the tutorial prose/snippets ahead of the example actually loaded by `<Go>`, so please keep this gated on that companion change landing.

The EN/ZH edits are otherwise aligned, CI is green, and `listHeight` itself is already described in the `<list>` API reference.

PR #79, opened 2025-03-16
Docs: Adds docs for downloading the lynx explorer from app store

pranshuchittora:
Change-Id: I76e4977d5b62f599717a40e7766b2557a5b3a8b0

Screenshot 2025-03-16 at 13 08 28

PR #1282, opened 2026-08-07
docs(product-detail): demonstrate the swiper in the web preview

Huxpro:
## Summary

The Product Detail tutorial teaches a touch-driven swiper, and its web previews are driven by a mouse. Those never produce `touchstart`/`touchmove`/`touchend`, so to a desktop reader every preview in the tutorial — including the finished swiper at the top — is a still image. The one thing the tutorial exists to demonstrate is invisible unless you scan the QR code.

This passes `autoGesture` to each preview. The swipe is replayed with synthesized touches, walking to the last image and back, with the contact point drawn the way a device simulator does it. The first trusted pointer or touch stops the playback, so a reader who reaches for the preview always wins.

`SwiperEmpty` is deliberately left alone: the prose introduces it *because* it does not scroll yet, and demonstrating a gesture that does nothing there would read as a broken preview rather than as the point being made.

A short note under "What Are We Building?" tells readers, in both locales, that the swipe is simulated and that touching the preview hands control back.

## Uses the released `@lynx-js/go-web` 0.9.0

`autoGesture` is a `<Go>` prop as of [go-web 0.9.0](https://www.npmjs.com/package/@lynx-js/go-web/v/0.9.0) (lynx-community/go-web#80, merged and published). This PR bumps `^0.8.1` → `^0.9.0` and passes the prop; nothing is vendored.

Playback is gated on the preview being on screen, inside go-web. This page carries eleven previews and the driver aims its contacts with `document.elementFromPoint`, which is viewport-relative and returns nothing below the fold — so without the gate ten previews would start a playback that silently hits nothing.

Safari is handled in go-web too: it refuses `new Touch()`, so the driver falls back to `document.createTouch` + `TouchList`.

## Test plan

- [x] `pnpm install` resolves `@lynx-js/go-web` 0.9.0.
- [x] Verified end-to-end against the running dev server on the **published package**, in headless **Chromium** and **WebKit**: 210 touch events delivered, the carousel travels a **1390px** range from ~10 `touchstart`s and ~330 `touchmove`s. Identical in both engines, no page errors.
- [x] Both locales carry the same 10 wired previews at parallel positions.

- Upgrade `@lynx-js/go-web` to `^0.9.0`.
- Configure `swipeDemo` for all Product Detail web previews in English and Chinese tutorials.
- Demonstrate looping bidirectional swipes until user interaction takes control.
- Document simulated swipe behavior, including `SwiperEmpty`, in both locales.


Huxpro:
Review update for `40fdd41b`: **not merge-ready yet**. The new local-driver revision removes the go-web release dependency, but I validated three correctness/lifecycle blockers in the current head:

- User handoff is not sticky across visibility changes: trusted input stops playback, but `attachAutoGesture()` clears `playingFor` off-screen, so scrolling away and back restarts the demo for the same page root. Preserve a per-page-root surrendered/opt-out state until refresh replaces that root.
- `steps: []` with the default `loop: true` enters the `do/while` with no `await` in its body and can synchronously spin forever. Guard the empty sequence before looping.
- `Go` attaches in `useEffect([])` while the wrapper is conditional on `autoGesture`; if that prop changes after mount, the driver can fail to attach or remain polling/observing a detached host. The attachment lifecycle should track enable/disable without restarting for merely recreated equivalent options.

Cross-repo consistency also needs care here: `auto-gesture.ts` says it is vendored upstream-first from lynx-community/go-web#79, but #79 is still open with unresolved findings in this same driver (including `stop()` clearing the active timeout without resolving the pending `wait()` promise). Please resolve/fix upstream and re-copy the reviewed version rather than landing a known-to-diverge snapshot.

CI is green and the EN/ZH tutorial changes are aligned; the blockers are limited to the new gesture-driver/lifecycle behavior.

Huxpro (commented):
The switch to released `@lynx-js/go-web@0.9.0` removes the vendoring/lifecycle issues from the previous revision, and CI is green, but I still see one user-facing behavior mismatch that makes this unsafe to merge under the conservative policy.

`go-web`'s current `WebIframe` implementation keeps a `playback` handle while the preview is visible, calls `playback.stop(); playback = null` when it leaves the viewport, and creates a fresh playback when it re-enters. A trusted user pointer/touch stops the inner `playAutoGesture`, but that opt-out state is not surfaced to `WebIframe`. So after the reader touches the preview, scrolls it off screen, and comes back, autoplay starts again for the same preview.

That contradicts the new tutorial copy (`Touch or click ... and it hands control straight back to you`) and the PR's stated "reader always wins" contract. Please fix this upstream-first in go-web (persist a surrendered-by-user state for the current preview/root across visibility changes), publish the follow-up, and bump here; or otherwise change the integration so a reader takeover cannot be silently undone by a scroll round trip. Once that is fixed, the website diff itself looks appropriately localized and EN/ZH-aligned.

Huxpro (commented):
Re-reviewed `7ecfb433`. The one new commit only wires `autoGesture` into `SwiperEmpty` (with matching EN/ZH explanatory copy); that change itself is coherent, but it does not resolve the existing takeover/lifecycle blocker.

The released `@lynx-js/go-web@0.9.0` still clears `playback` when the preview leaves the viewport and calls `playAutoGesture(...)` again when it re-enters. `playAutoGesture` stops on a trusted user event, but that user-surrender state is local to the discarded handle and is not persisted by `WebIframe`. So a reader can click/touch to take control, scroll away, then return and have autoplay resume for the same preview.

That still conflicts with the new tutorial promise that interaction “hands control straight back to you” / “控制权会立刻交还给你” and the PR’s “reader always wins” contract. CI is green, and the new SwiperEmpty demonstration is fine, but under the conservative merge policy I would keep this blocked until go-web persists the user opt-out across visibility changes (upstream-first, then release/bump here), or the website integration/copy is changed so the behavior and contract actually match.

Huxpro:
Thanks — these all landed on the vendored copy, which this PR now deletes: the site takes `autoGesture` from the published `@lynx-js/go-web` 0.9.0 instead, so `auto-gesture.ts`, `attach-auto-gesture.ts` and the wrapper's own prop are gone. The two Copilot notes about `Go.tsx` are moot for the same reason — it is a plain pass-through again.

Two of them are real upstream, though, and I am carrying them to go-web rather than dropping them:

- **Empty `steps` spins the main thread.** Confirmed: with `steps: []` and `loop: true` nothing inside the `do`/`while` ever awaits, so `stopped` can never be set by a timer or event. That code is now published, so it needs a patch release.
- **A scroll round trip restarts playback after the reader took over.** The same shape exists in go-web's own `IntersectionObserver` gate. Worth deciding there, since telling the two cases apart needs `playAutoGesture` to report *why* it stopped.

Resolving here because the files they annotate no longer exist on this branch.

PR #1283, opened 2026-08-09
feat(home): add Octane to Framework Agnostic, refresh the feature card copy

Huxpro:
## What

Adds **Octane** to the homepage's Framework Agnostic card and rewrites the copy on three feature cards (EN + ZH).

### Framework Agnostic

- New **Octane** entry between Vue and Miso, linking to <https://octanejs.dev>.
- Miso's button reads `Miso` instead of `Miso (Haskell)`.
- Copy reframed from "Lynx isn't limited to React" to Lynx being a platform for frameworks. Every framework integrates through **the same standardized APIs**, which is what carries the point that being the official flavor buys no privileged path. ReactLynx sits inside the list rather than ahead of it, and the card closes on "All frameworks welcome."

### Performance at Scale

- The multithreaded story now spans **engine and framework**, not just the engine, hinting at the patterns the architecture opens up rather than naming them.

### Web-Inspired Design

- `CSS and React` → `CSS and JavaScript`, and `web knowledge` → `the web's programming model`, so the card describes the web platform rather than one framework on it. This also gives the EN copy the paradigm framing the ZH already had in 延续 Web 开发范式.

## The Octane mark

`docs/public/assets/home/octane-logo.svg` is the flame mark from the octanejs.dev wordmark. The outer shape is `fill="currentColor"` (as it is on their site), so it tracks the button text color in both light and dark themes; the inner flame keeps the `#FF415A` brand red. `.octane-icon` pins the color to `var(--home-blog-btn-color)`, matching how the ReactLynx icon is handled.

Rendered at 17×20 (viewBox ratio 0.857), next to Vue's 20×20 and Miso's 18×18.

## Layout constraints

Both rewritten cards are height-constrained, so the copy was measured rather than eyeballed — rendered at the real column width (378px), 14px, line-height 1.5 in headless Chromium:

- **Performance at Scale** sits above a 192px visual, leaving ~95px for the description (400px card − 72px padding − 41px title − 192px). 4 lines is 100px and would clip under `overflow: hidden`; the shipped copy is 3 lines / 79px. `perf-driven` rather than `performance-driven` is what buys that.
- **Web-Inspired Design**'s Moon is `position: absolute` at ≥1015px, so the box is looser, but copy past 3 lines starts to collide with the ring visually. Shipped EN is 3 lines, ZH 2 — same as before.

## Testing

- `tsc --noEmit` — no errors in `src/components/home-comps/features/`
- `prettier --check` and `scripts/ci/check-asset-urls.mjs` pass on the touched files
- Rendered the new SVG in headless Chromium to confirm the mark and its two-tone fill come out correctly at button size

> Two environment notes: the pre-commit hook could not run here — `pnpm install`'s `prepare` script shells out to `rsync`, which isn't present in the container — so the hook's actual checks (prettier, check-asset-urls) were run manually as listed above. And the three failing Cloudflare Pages checks are pre-existing: #1281 shows the identical three failures on an unrelated branch. Netlify's preview builds fine.

- Add Octane to the Framework Agnostic card with a link to `octanejs.dev`.
- Update the Miso label to `Miso`.
- Rewrite Framework Agnostic, Performance at Scale, and Web-Inspired Design copy in English and Chinese.
- Add a theme-aware Octane flame SVG icon.
- Verify layout constraints, TypeScript, formatting, asset URLs, and SVG rendering.


Huxpro (commented):
Blocking before this is ready:

1. **Octane needs an explicit maturity/compatibility qualifier.** Octane's current Lynx docs label the integration **Early access** and say `@octanejs/lynx` / `@octanejs/rspeedy-plugin` are not published yet. Its renderer README is even more explicit: the package is `0.0.0`, private, not a native renderer release, with Android/iOS and other native evidence still open. Presenting Octane as a peer framework on the Lynx homepage without any status/version metadata overstates the current support level. Please either defer the listing until there is a public release, or surface an Experimental/Early access qualifier and link directly to the canonical Lynx integration/status docs; once published, include the package/engine version floor as well.

2. **`all integrate through the same standardized APIs` is stronger than the current contract.** Octane currently describes its universal renderer ABI as experimental and says it follows a pinned Lynx/ReactLynx interoperability boundary. Unless there is a canonical Lynx framework-integration API/spec we can point to, please use wording that does not imply a stabilized standardized framework API contract.

3. **Please keep the PR scoped to the framework-card change.** The additional rewrites to the Performance and Web Inspired cards introduce separate technical/positioning claims (notably `The same architecture lets data load while the UI renders`) that are unrelated to adding Octane/Miso and make cross-site consistency harder to review. I'd split/revert those from this PR unless they have their own source-of-truth alignment audit.

The EN/ZH framework-card edits themselves are kept aligned, and the logo/action wiring looks structurally straightforward; the concerns above are about compatibility/status accuracy and scope, so I don't think this should merge in its current draft.

Huxpro (commented):
Re-reviewed current head `6abcdd9`. The latest commit does remove the earlier `data load while the UI renders` claim, so that specific concern is addressed. I still don't think this is merge-ready:

- The framework card still presents Octane as an unqualified peer. Octane's current canonical Lynx page labels the integration **Early access**, says `@octanejs/lynx` and `@octanejs/rspeedy-plugin` are not published yet, and explicitly says not to use it for a production app yet. Its current `packages/lynx/status.json` keeps both packages private at `0.0.0` and says the private packages are not a technical preview. Please surface that maturity/compatibility status (or defer the listing until there is a public release), with the relevant version floor once one exists.
- `all integrate through the same standardized APIs` is unchanged and still implies a stabilized common framework API contract that the current Octane integration does not claim. Please soften this unless there is a canonical Lynx framework-integration spec/API to link.
- The Performance and Web Inspired card rewrites are still unrelated to the Octane/Miso change. The new Performance wording is less problematic than the previous sentence, but these remain separate homepage positioning changes and should be split/reverted for a low-risk framework-card PR.

CI is green on this head, and the EN/ZH edits remain aligned, but the compatibility/status and scope issues above are substantive blockers.

Huxpro:
/cherry-pick release/3.9, release/4.0


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1287, opened 2026-08-10
docs(css): document the :hover pseudo-class

AdrianLCA:
## Summary

- Add `:hover` to the supported pseudo-class list in the English and Chinese CSS selector references.
- Clarify that `:hover` depends on host pointer hover input and is not activated by focus navigation.

## Checks

- `prettier --check` on both changed files
- `cspell lint` on both changed files
- `node scripts/ci/check-api-summary.mjs` on both changed files
- `node scripts/ci/check-asset-urls.mjs` on both changed files
- `zhlint docs/zh/api/css/selectors.mdx`
- `python3 -m scripts.check_api_doc.index`
- `git diff --check`

- Document `:hover` behavior in the English and Chinese CSS selector references.
- List `:hover` among supported pseudo-classes.
- Define `:hover` compatibility metadata and pointer-hover requirements.

### DIFF
--- a/docs/en/api/css/selectors.mdx
+++ b/docs/en/api/css/selectors.mdx
@@ -1,3 +1,9 @@
+---
+api: css/selectors.hover
+---
+
+import { APISummary, APITable } from '@lynx';
+
 # Selectors
 
 CSS selectors specify which elements a CSS rule applies to. Pedantically speaking, CSS selectors is an umbrella term that includes Selectors, Combinators, Pseudo-elements, and Pseudo-classes.
@@ -234,6 +240,7 @@ A `:` pseudo-class is a type of selector used to select elements in a specific s
 Supports
 
 - `:active`
+- `:hover`
 - `:not()`
 - `:root`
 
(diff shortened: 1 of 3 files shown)


Huxpro (changes requested):
Blocking on compatibility/version metadata. This PR adds `:hover` as a supported CSS pseudo-class, but the new text only says support “depends on the platform and renderer” and does not tell readers which targets actually support it or from which LynxSDK/engine version. That is material for a pointer-dependent feature and is especially important here because the branch itself is named `docs/4.2-hover-pseudo-class`; if 4.2 is the intended floor, it needs to be visible in the documentation/compatibility source rather than implicit in the branch name. If 4.2 is not the floor, please document the actual first supported version.

Please add an actionable compatibility statement/matrix (and the repository’s compatibility metadata/badges if this API category supports them) covering the supported native/desktop/Web renderers/platforms and the first supported version. The upstream `lynx-family/lynx` API skill already documents `:hover` but likewise only says that support varies by platform, so copying that ambiguity into the canonical website does not resolve the compatibility question.

The EN/ZH changes themselves are synchronized and the basic semantics look fine. I’m not merging until the compatibility boundary is explicit; CI/CLA are also still pending on this head.

Huxpro (approved):
The compatibility blocker is resolved on this head. The new EN/ZH docs are aligned and now carry structured compatibility metadata plus an explicit matrix. I cross-checked the Clay 3.7 source: `BaseView` wires mouse enter/leave into `OnHoverChanged`, the Lynx event dispatcher maps that to `kPseudoStateHover`, and the Clay build config enables mouse tracking on macOS/Windows and full Android builds while excluding iOS and Clay Lite. The classic Android tree exposes the pseudo-state constant but has no corresponding hover activation path. Web is correctly left as browser-native support with an unknown exact package floor (`version_added: true`). Compat-data stats and CI are green. This is additive documentation/data only and looks low-risk.

PR #1288, opened 2026-08-11
docs(elements): document Clay scrollbar attributes and semantics

icecreamx10:
## Summary

The Website only documented Clay scrollbar controls in fragments, and the handwritten/generated pages disagreed on the enable attribute's name, default, and behavior.

This PR is intentionally scoped to Clay. It:

- documents the non-deprecated `enable-scrollbar` Clay attribute for `scroll-view` and `list`;
- records Lynx `3.7` as the added version for all documented Clay scrollbar attributes, based on the first public release tag containing the implementation;
- clarifies that enabling a Clay scrollbar uses auto-hide by default (`scroll-bar-auto-hide=true`, `scroll-bar-auto-hide-delay=1000`) and that `scroll-bar-auto-hide=false` keeps it visible while content overflows;
- documents the Clay-only delay, sizing, thumb, track, color, hover, and active attributes with implementation defaults;
- aligns English/Chinese and generated/handwritten references;
- adds Clay-only `scroll-view`/`list` compatibility data, regenerated stats, and tests that prevent these attributes from being presented as cross-platform;
- leaves the existing deprecated `scroll-bar-enable` documentation untouched and does not add it to compatibility data;
- does not add Web-specific scrollbar spellings or the proposed `none | auto | persistent/always` enum.

## Public implementation references

- [Clay scrollbar attribute handling and shared wrapper](https://github.com/lynx-family/lynx/blob/2e38a604713f5e632a55c207bd452f1116134474/clay/ui/component/scrollbar/scrollbar_wrapper.cc#L23-L95)
- [Clay auto-hide and visual defaults](https://github.com/lynx-family/lynx/blob/2e38a604713f5e632a55c207bd452f1116134474/clay/ui/component/scrollbar/scrollbar_view.h#L77-L91)
- [Lynx 3.7.0 Clay scrollbar implementation](https://github.com/lynx-family/lynx/blob/3.7.0/clay/ui/component/scrollbar/scrollbar_wrapper.cc#L19-L97)

## Validation

- `pnpm --filter @lynx-js/lynx-compat-data run lint`
- `pnpm --filter @lynx-js/lynx-compat-data run lint:keys`
- `pnpm --filter @lynx-js/lynx-compat-data run test` (21 tests passed)
- `pnpm --filter @lynx-js/lynx-compat-data run gen-stats`
- `pnpm run check-docs`
- `node scripts/ci/check-api-summary.mjs`
- `pnpm run format:check`
- `pnpm run build`
- pre-commit cspell and asset URL checks

`pnpm run zhlint` remains blocked by an existing `zhlint` TypeError in unrelated Lynxtron docs; targeted runs only reported pre-existing lines outside this change.


Huxpro (changes requested):
Blocking this for API-source and compatibility consistency.

1. `docs/*/api/elements/built-in/scroll-view-API.mdx` is a TypeDoc-generated API surface. The website's own contributor docs explicitly say not to edit generated API files directly because they are overwritten; the source comments/types need to be updated and TypeDoc regenerated. The current upstream `ScrollViewProps` still exposes only `scroll-bar-enable`, says `@defaultValue true` / `@since 1.4`, and does not contain `enable-scrollbar` or the new Clay scrollbar styling attributes. `ListProps` likewise exposes only `scroll-bar-enable` and documents different default semantics. Landing this as-is would immediately leave the typed/source API and website reference out of sync and risks the changes disappearing on regeneration.

2. The new compatibility records are largely unversioned (`version_added: true`) for Clay/Web and for several native list entries. Please record the first supported LynxSDK/renderer version per platform where it is knowable, rather than only marking support as present. This is especially important for the newly surfaced Clay-only attributes and aliases.

3. `scroll-bar-enable` now shows Android/iOS/Clay/Harmony/Web badges followed by a single `<VersionBadge>1.4</VersionBadge>`, while the structured data only assigns `1.4` to iOS and leaves the other platforms at different/unknown floors. That badge reads as a cross-platform 1.4 floor and is misleading. Please make the version presentation platform-specific or rely on the compatibility table once the per-platform versions are populated.

The implementation/default corrections themselves look directionally sound (for example Android and iOS scroll-view reset the native indicators to false, and Clay's wrapper accepts both spellings), but the public source-of-truth and version metadata need to be aligned before this is safe to merge.

Huxpro (changes requested):
Re-reviewed the updated head. The compatibility/version issues from the previous pass are now addressed well: the new Clay-only attributes have explicit `3.7` badges and `version_added: "3.7"` entries for all Clay targets, and that floor is consistent with the public implementation appearing in the 3.7.0 tag (the same scrollbar wrapper path is absent from 3.6.0).

One substantive blocker remains: this PR still directly edits `docs/{en,zh}/api/elements/built-in/scroll-view-API.mdx`, while the website contributor guide says TypeDoc-generated API files must not be edited manually because they will be overwritten. More importantly, the current upstream `ScrollViewProps` source still exposes only `scroll-bar-enable`; it does not declare `enable-scrollbar` or any of the new `scroll-bar-auto-hide` / sizing / color attributes. So the website would advertise properties that the canonical TypeScript API source does not yet expose, and regeneration can erase this documentation.

Please land/update the source type/comments first (or otherwise establish the correct source-of-truth path) and regenerate the API reference from that source. Once that source/reference mismatch is resolved, the remaining Clay-scoped documentation and compatibility metadata look mergeable.

icecreamx10:
感谢 review。重新核对后，我们将公开 API 的依据收紧为 npm latest 发布包。当前最新正式版本 @lynx-js/types@4.1.0 仅包含既有的 scroll-bar-enable；enable-scrollbar 以及 auto-hide、尺寸和颜色等新增 scrollbar 属性尚未进入发布包。因此本 PR 已撤回全部相关文档、compat 数据和测试变更，当前 diff 为 0。等这些类型随正式版本发布后，再从发布包的 TypeScript source-of-truth 通过 codegen 生成 API 文档并重新提交。先关闭本 PR。

PR #1290, opened 2026-08-11
fix: detect ui subsite under 4.0 base

fzx2666-fz:
## Summary
- normalize the pathname with removeBase before matching subsites
- fix SSR subsite detection for version-prefixed routes such as /4.0/zh/ui/

## Validation
- pnpm run build
- verified doc_build/zh/ui/index.html renders with data-subsite="ui"

### DIFF
--- a/theme/index.tsx
+++ b/theme/index.tsx
@@ -50,7 +50,7 @@ const TYPING_START_DELAY_MS = 1600;
 
 // Match subsite by checking if any path segment exactly equals the subsite value
 const findSubsite = (pathname: string) => {
-  const segments = pathname.split('/');
+  const segments = removeBase(pathname).split('/');
   return SUBSITES_CONFIG.find((s) => {
     if (s.value === 'ui') {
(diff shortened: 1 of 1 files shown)


Huxpro (changes requested):
The patch does not appear to change the behavior described in the PR, so I don't think this is safe to merge as a bug fix yet.

On `release/4.0`, Rspress `base` comes from `SITE_BASE`, and `docs/public/version.json` currently gives 4.0 `docs_link: "/"`; with that base, `removeBase(pathname)` is effectively an identity operation. Even if this branch later uses `/4.0` as its base, the existing implementation already splits `/4.0/zh/ui/` into path segments containing the exact `ui` segment, so `findSubsite()` should already return the `ui` subsite before this change.

Please add a regression test or otherwise show a concrete pathname/base pair that fails on the parent commit and passes with this patch, then make the fix target that actual failure. The current validation only checks the post-change output, which is also consistent with the pre-change logic.

PR #1293, opened 2026-08-11
fix(spec): flush HTML viewer to edges, add open-externally button

Huxpro:
## Summary
- Remove gray iframe background gutters so the living spec content sits flush against the content area edges
- Consolidate redundant per-breakpoint media queries into a single rule
- Add a hover-reveal "open in new tab" button (top-right corner) so users can view the spec standalone

## Test plan
- [ ] Verify spec page at `/guide/spec.html` has no visible gray borders around iframe
- [ ] Hover over the content area and confirm the external-link button appears top-right
- [ ] Click the button to confirm it opens the living spec in a new tab
- [ ] Test in dark mode

- Remove iframe gutters and preserve full-width spec content.
- Consolidate responsive `rspress-doc` styles.
- Provide a hover-revealed control to open the living spec in a new tab.
- Support light and dark mode control styling.
- Add rounded corners, bottom spacing, and a shadow to the spec viewer.

### DIFF
--- a/src/components/html-viewer/index.module.less
+++ b/src/components/html-viewer/index.module.less
@@ -36,7 +36,9 @@
     box-shadow: 0 1px 4px rgba(0, 0, 0, 0.12);
     backdrop-filter: blur(4px);
     opacity: 0;
-    transition: opacity 0.2s ease, background 0.15s ease;
+    transition:
+      opacity 0.2s ease,
+      background 0.15s ease;
     text-decoration: none;
   }
 
--- a/src/components/html-viewer/index.module.less
+++ b/src/components/html-viewer/index.module.less
@@ -1,7 +1,9 @@
 .html-viewer-frame {
   position: relative;
-  width: 100%;
-  height: calc(100vh - 72px);
+  margin: calc(-1 * var(--rp-content-padding-y))
+    calc(-1 * var(--rp-content-padding-x));
+  width: calc(100% + 2 * var(--rp-content-padding-x));
+  height: calc(100vh - var(--rp-nav-height, 64px));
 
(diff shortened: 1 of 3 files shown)


Huxpro (commented):
Blocking issues on this head before merge:

- The new external-link control is `opacity: 0` but remains focusable/clickable, and it is revealed only by mouse hover. Keyboard users can tab to an invisible control, and the hidden 32×32 anchor can intercept pointer input. Please reveal it for keyboard focus (`:focus-within`/`:focus-visible`) and disable pointer events while hidden.
- The new-tab link uses only `formattedPath`, while the displayed iframe includes `location.hash`; after navigating within the living spec, "open in new tab" therefore loses the currently viewed anchored section. Keep the external URL aligned with the iframe/current hash.
- CI on this exact head is red in Format Check, with Build skipped, so the change has not passed the repository's merge gate.

These are small fixes, but under the low-risk merge bar this head should not be merged yet.

Huxpro (commented):
Re-reviewed current head `2147a6d`. The new commits add the bottom spacing/rounded-shadow treatment and fix Prettier, but the existing merge blockers are still present on the current lines:

- `.open-external` is still `opacity: 0` and revealed only via `:hover`, so keyboard focus remains invisible and the hidden anchor still participates in hit-testing.
- the external link still uses only `formattedPath`, while the iframe URL includes `location.hash`, so opening externally loses the currently viewed anchored section.

The existing inline threads for both issues are still unresolved/current, so I’m not duplicating them inline. Format Check is now green on this head; the CI Build step is still running. Keeping this blocked until the two functional/accessibility issues are fixed and CI completes.

Issue #1291, opened 2026-08-11
Document MainThreadObject in the ReactLynx/MTS guide


Huxpro:
Tracks the documentation deliverable from https://github.com/lynx-family/lynx-stack/issues/3446.

Depends on the accepted API from https://github.com/lynx-family/lynx-stack/issues/3447 and uses the Motion and Element integrations from https://github.com/lynx-family/lynx-stack/issues/3448 and https://github.com/lynx-family/lynx-stack/issues/3449 as worked examples.

## Outcome

Publish one focused ReactLynx/MTS documentation chapter explaining `MainThreadObject`, its handle representation, and its relationship to `MainThreadRef` and `MainThread.Element`.

## Audience

- ReactLynx application developers using main-thread functions;
- library authors integrating stateful main-thread object models such as MotionValue;
- framework contributors working on worklet capture and hydration.

### Mental model

- background handle vs realized main-thread object;
- `MainThreadObject`, `MainThreadObjectType<I, O>`, and `MainThreadObjectHandle<O>`;
- constructed stable objects vs renderer-controlled bindings;
- pure-JS vs native-backed objects.

### Existing primitives

- `MainThreadRef<T>` as a mutable cell/binding;
- MotionValue as a factory-realized pure-JS main-thread object;
- `MainThread.Element` as a native-backed JavaScript object model;
- `main-thread:ref` as a rebindable `MainThreadRef<MainThread.Element | null>`.

### Rules and non-goals

- methods are callable only inside main-thread functions;
- initialization payloads must be serializable;
- main-thread mutations do not rerender React or synchronize back to background;
- handles are not Proxies, RPC stubs, shared values, or transferable objects;
- Element lifecycle remains renderer-owned;
- when to prefer `useMainThreadRef` over a library-provided object hook.

### Examples

- a minimal library-author object type definition;
- `useMotionValue` usage;
- `main-thread:ref` / `MainThread.Element` usage;
- one misuse example with the expected diagnostic.

## Deliverable

A single PR to `lynx-family/lynx-website`, ideally one new chapter under the existing ReactLynx main-thread script / MTS documentation rather than several disconnected pages.

## Acceptance criteria

- [ ] Documentation matches the accepted core API and terminology.
- [ ] One chapter covers both the conceptual model and practical usage.
- [ ] MotionValue and Element examples make ownership/binding differences explicit.
- [ ] All code samples build or are validated against the implementation PR.
- [ ] Site checks and link checks pass.
[...]

PR #1294, opened 2026-08-11
docs(react): explain MainThreadObject

Huxpro:
## Summary

Tracks lynx-family/lynx-stack#3446.
Implements #1291.

Implementation references:
- Core API and protocol: lynx-family/lynx-stack#3453
- Motion migration: lynx-family/lynx-stack#3455
- MainThreadRef / Element integration: lynx-family/lynx-stack#3456

This adds one focused ReactLynx/MTS chapter, localized at the same route in English and Chinese, that explains:

- background handles versus stable realized main-thread objects;
- `MainThreadObjectType<I, O>`, `MainThreadObjectHandle<O>`, and `useMainThreadObject()`;
- library-owned pure-JS objects versus renderer-owned native-backed Element bindings;
- `MainThreadRef`, `useMotionValue`, `main-thread:ref`, and `MainThread.Element`;
- serialization, thread-access, mutation, ownership, and cleanup rules;
- the exact development diagnostic for background-runtime misuse.

It also adds the chapter to both ReactLynx sidebars and updates the existing Motion capability table now that `useMotionValue()` is supported.

Published routes after merge:
- `/react/main-thread-object`
- `/zh/react/main-thread-object`

## Sample validation

The Counter factory, `useMotionValue`, and `main-thread:ref` examples were combined into a temporary TSX fixture and strictly typechecked against the built artifacts from lynx-family/lynx-stack#3455 plus `@lynx-js/types` 4.1. The fixture passed and was removed; it is not part of this PR.

This validation caught and corrected the event type to the current `MainThread.TouchEvent`.

## Site validation

Passed:

- full Rspress production build
  - 3,181 sitemap pages generated
  - English and Chinese MainThreadObject HTML/Markdown routes generated
  - Rspress dead-link checking passed
- full-repository Prettier check
- exact lint-staged/pre-commit task set on all six changed files
  - Prettier
  - CSpell
  - API-summary check
  - asset URL check
- CSpell on both new chapters and both touched Motion pages
- zhlint on the Chinese chapter
- case-collision check
- full asset URL check
- API documentation consistency check across 1,717 files
- `git diff --check`

The build emitted one existing warning from `vscode-languageserver-types` using a dynamic `require`.

## Environment and unavailable check

A first isolated-worktree build reached SSG but could not read gitignored generated example metadata under `docs/public/lynx-examples`. Reusing the matching repository's generated baseline artifacts allowed the full build above to complete; no generated artifacts are committed.

This repository has no dedicated lint or typecheck package script. As an additional check, `tsc --noEmit -p tsconfig.json` was attempted. It is not currently clean on `main`: it reports existing errors in `rspress.config.ts`, React/Rspress dependency type skew, missing optional test/editor dependencies, i18n key generation, and compatibility-data types. None point to the changed MDX files. The code samples were therefore validated independently against the implementation artifacts as described above.

Go/no-go: **GO for review**. No documentation-specific blocker remains.

### DIFF
--- a/docs/en/react/_meta.json
+++ b/docs/en/react/_meta.json
@@ -7,6 +7,7 @@
   "thinking-in-reactlynx",
   "lifecycle",
   "main-thread-script",
+  "main-thread-object",
   "code-splitting",
   "react-compiler",
   {
--- a/docs/en/react/main-thread-object.mdx
+++ b/docs/en/react/main-thread-object.mdx
@@ -1,0 +1,187 @@
+---
+description: 'Understand MainThreadObject handles and choose between library-owned objects, MainThreadRef, and renderer-owned MainThread.Element bindings.'
+---
+
+# Main-Thread Objects
+
+[Main thread functions](./main-thread-script) run in a different JavaScript runtime from your React component. Captured JSON values are copied into that runtime. A stateful object, however, needs identity: every main thread function that captures it should reach the same object.
+
+`MainThreadObject` is the ReactLynx model for that identity. A hook returns an opaque background-runtime handle. When a main thread function captures the handle, ReactLynx resolves it to one stable object in the main-thread runtime.
+
+```text
+background runtime                 main-thread runtime
+
+opaque handle ── capture/hydrate ──> stable realized object
+     │                                      │
+     └─ identity + initialization data      └─ methods and mutable state
+```
+
+The handle is not the realized object. It is not a Proxy that forwards operations, an RPC stub, a shared value, or a transferable JavaScript object. Its serialized representation is framework-owned metadata; do not inspect it or send it through an application protocol.
+
+## The Core Types
+
+A library describes a main-thread object with three core APIs:
+
+- `MainThreadObjectType<I, O>` maps a JSON-serializable initialization payload `I` to a realized object `O`. It contains a globally unique, stable `type` key, a `create` function, and an optional `dispose` function.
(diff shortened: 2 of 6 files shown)


Huxpro (commented):
Documentation blockers before this can land:

- This chapter presents `defineMainThreadObjectType`, `useMainThreadObject`, `MainThreadObjectType`, and `MainThreadObjectHandle` as current public APIs, but the implementation they depend on is still in draft/open lynx-stack PR #3453, and the Motion migration/example is still in draft/open #3455. The website should not publish these as stable current APIs before the implementation/release boundary is known; add explicit first-supported/unreleased version metadata and align landing order with the implementation.
- The PR introduces those public APIs without corresponding API-reference coverage. The current website has no API-reference entries for the MainThreadObject APIs, and the new chapter only links the pre-existing `MainThreadRef`, `useMainThreadRef`, and `MainThread.Element` references. Please ensure the generated React API reference contains and links the new types/functions before publishing the guide.
- The Motion capability table changes `useMotionValue` from "not yet supported" to supported while leaving the existing `Lynx Engine >= 3.5` / `ReactLynx >= 0.115.2` minimums unchanged. The newly documented behavior depends on the new MainThreadObject runtime/package changes, so those old compatibility floors are not established for `useMotionValue`. Please record the actual first-supported package/runtime versions (or clearly mark the row as unreleased) so the table does not imply support on older versions.

EN/ZH parity in this PR looks good; the blockers are release/version/API-reference alignment.

PR #1305, opened 2026-08-14
docs(lynxtron): document windowless LynxWindow option

FrendyChen:
Add the windowless option to LynxWindowConstructorOptions in both English and Chinese API docs, including Linux headless-only behavior and supported use cases

- Document `deviceScaleFactor` and `windowless` in English and Chinese `LynxWindowConstructorOptions` API references.
- Describe Linux headless-only behavior and supported server-side use cases.
- Specify `deviceScaleFactor` defaults and `windowless` support since v0.0.4.
- Correct `Defined in` links to `lynx-window.d.ts`.


Huxpro (changes requested):
Thanks — the behavior is useful to document, but I don’t think this is safe to land in its current form.

The two files changed here explicitly say they are generated by `@lynx-js/tool-typedoc` and “Do not edit this file directly.” Please put the documentation/compatibility information in the canonical `lynxtron` TSDoc source and regenerate the EN/ZH reference rather than hand-editing generated output.

There is also a source-of-truth completeness issue: the current `LynxWindowConstructorOptions` source contains both `deviceScaleFactor` and `windowless`, and they were introduced together by the windowless feature commit. This PR manually adds only `windowless`, leaving the generated API reference incomplete. Regenerating from the canonical source should keep the full public surface aligned.

Please also add the release boundary. The windowless feature commit is after `v0.0.3` and is an ancestor of `v0.0.4`, so the public option is available from `v0.0.4`; the API reference should make that first-supported version explicit instead of presenting it as timeless. The Linux headless-only statement should likewise live in / be verified by the canonical TSDoc source so the generated docs cannot drift.

Finally, the current generated block says `lynx-window.d.ts:248` while linking to `#L37`; regeneration should fix that stale source-location label too.

FrendyChen:
Thanks, good catch. I’ve addressed the issues in this update:

- documented both `deviceScaleFactor` and `windowless` together in EN/ZH
- added the release boundary (`supported since v0.0.10`)
- fixed the stale `Defined in` source links so they point to the correct lines in `lynx-window.d.ts`

Please take another look.

Huxpro (changes requested):
Thanks for updating this. I re-checked the current head against the canonical Lynxtron source and release tags, and there are still source-of-truth/version correctness issues that make this unsafe to merge as-is.

1. These files are generated TypeDoc output; the availability/default metadata should live in the canonical `lynx-window.d.ts` TSDoc (or its generation metadata) and the EN/ZH reference should be regenerated, rather than hand-edited here.
2. The new `deviceScaleFactor` default in this PR is inconsistent with the canonical source: `LynxWindowOptions.deviceScaleFactor` documents the default as `1`, not `2`.
3. The `0.0.10` first-supported/version claim does not match the public release history. Both `deviceScaleFactor` and `windowless` are already present in the tagged `v0.0.4` public declaration (and also in `v0.0.9`), so `0.0.10` cannot be their introduction floor.

Please derive the exact `@since`/availability boundary from the upstream implementation/release history, correct the canonical source metadata/default, and regenerate both language references from that source.

FrendyChen:
Thanks for the detailed re-check. I updated the canonical `lynx-window.d.ts` TSDoc for `deviceScaleFactor` so its default value is documented consistently, then aligned the EN/ZH API references with that TSDoc.

I also corrected the availability boundary to `v0.0.4`, since both `deviceScaleFactor` and `windowless` are already present in the tagged `v0.0.4` public declaration, and updated the generated reference source links to match the current canonical file locations.

Please take another look.

Huxpro (changes requested):
Thanks for updating the canonical Lynxtron source and correcting the option-introduction floor to v0.0.4. I re-checked this exact head against the current source and tagged releases; two release/source-of-truth issues still make it unsafe to merge as-is.

1. These website files are generated TypeDoc output, but the new `Supported since v0.0.4` text still does not exist in the canonical `lynx-window.d.ts` TSDoc/generation metadata. A future regeneration would therefore drop that availability metadata. Please put the release boundary in the source-of-truth (or the generator's version metadata) and regenerate EN/ZH rather than preserving it only as a manual edit to generated files.

2. `deviceScaleFactor` now says `Default is 2. Supported since v0.0.4`, but that combines two different version histories. The option exists in v0.0.4, but the tagged v0.0.4, v0.0.10, and v0.0.15 declarations all document the default as `1`; only current `main` now says `2`. Please identify the first released version where the default changes to 2, or explicitly mark 2 as an unreleased/current-main default. Otherwise the page will tell users of released versions that their default is 2 when it is actually 1.

Also, the GitHub CI workflow for this head completed as `action_required` without running jobs, so there is no normal CI validation on `3b004584` yet.

FrendyChen:
Thanks for re-checking. I updated the canonical `lynx-window.d.ts` TSDoc and aligned the EN/ZH TypeDoc output with the latest source.

For `deviceScaleFactor`, I also re-checked the implementation history: the actual default value changed to `2` in `v0.0.10`, while the `lynx-window.d.ts` TSDoc was only updated in a later version after `v0.0.15`. So the docs now follow the actual runtime behavior rather than the TSDoc in version `v0.0.10` and `v0.0.15`.

FrendyChen:
I also checked the PR checks. Netlify, CLA, and CodeRabbit are green on the latest head, but the GitHub Actions suite still shows `action_required` with no jobs created. I do not see an `Approve and run` button on my side.

Huxpro (approved):
Re-reviewed current head `7f3fecb0`. The previous source/version blockers are resolved. Canonical Lynxtron TSDoc now owns the `0.0.4` availability metadata and Linux headless-only semantics, the generated source links match the current declarations, and implementation history confirms `deviceScaleFactor` defaults to `1` through v0.0.9 and changes to `2` in v0.0.10. The EN/ZH docs are aligned and the content looks correct/additive.

Approving the documentation content. I am not merging this head yet because the repository CI workflow concluded `action_required` rather than running the normal validation jobs; Netlify, CLA, and CodeRabbit are green.

Huxpro (approved):
Re-reviewed current head `d1ccab3b`. The PR diff is unchanged from the previously approved documentation content; the branch has only advanced by bringing in current `main`. The EN/ZH version/default metadata remains aligned and the current Netlify preview built successfully. Approving the current head. I am still not merging it because the repository `CI` workflow for this exact SHA concluded `action_required` instead of executing the normal jobs.

PR #1311, opened 2026-08-17
WIP: document custom framework Performance API integration

loongliu:
- Document `PipelineOptions` and `FlushOptions` APIs.
- Document framework performance hooks and timing keys.
- Provide custom framework Performance API integration guidance.
- Add integration navigation and API metadata.
- Track open decisions for the stable public contract.

### DIFF
--- a/CUSTOM_FRAMEWORK_PERFORMANCE_API_OPEN_ISSUES.md
+++ b/CUSTOM_FRAMEWORK_PERFORMANCE_API_OPEN_ISSUES.md
@@ -1,0 +1,117 @@
+# Custom Framework Performance API Documentation: Open Issues
+
+This file tracks decisions that must be resolved before the custom-framework Performance API integration is declared a stable public contract.
+
+The current documentation structure is:
+
+- `PipelineOptions`: `docs/en/api/lynx-api/performance-api/pipeline-options.mdx`
+- `__FlushElementTree`: `docs/en/api/engine/element-api/__FlushElementTree.mdx`
+- framework performance hooks: `docs/en/api/lynx-api/lynx/lynx-performance/`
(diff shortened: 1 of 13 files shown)


Huxpro (changes requested):
Blocking this head from merge. The PR itself records that the framework-facing contract is not yet stable: source-of-truth ownership, `PipelineOptions` fields, `_onPipelineStart` signatures, the public `FlushOptions` subset/overloads, cancellation semantics, allowed origins, timing-stage semantics, compatibility data, and publication status are all still open decisions. Those are normative API questions, so the API reference should not present one concrete contract until they are resolved in the canonical declarations.

In particular:

- Please add exact first-supported version/platform compatibility metadata (and the corresponding API badges/tables) for `PipelineOptions`, `__FlushElementTree`, and every underscore-prefixed framework hook. A prose warning that the flow was validated on Lynx 4.0.1 Android/iOS is useful, but it does not replace machine-readable compatibility or minimum-version metadata.
- Keep the canonical Performance API pages internally consistent. The new `PipelineOptions.stage` text defines framework-neutral `hydrate`/`update` semantics, while the existing `FrameworkPipelineTiming.stage` page still defines them in ReactLynx-specific terms (`hydrate` as BTS→MTS calibration and `update` as `useEffect` updates). Resolve that source-of-truth conflict rather than documenting two contracts.
- Add the corresponding Chinese API references, guide, and navigation updates. The existing Performance API surface is bilingual, and publishing only the EN half creates immediate cross-site drift.
- Prefer generating/deriving these framework-facing types from the agreed canonical package once CFP-01 is resolved, rather than freezing hand-authored declarations while the runtime declarations still disagree.

CI is green, but these are API-contract, compatibility, and cross-site consistency blockers rather than build issues.

PR #1327, opened 2026-08-19
fix(theme): decouple shared OG config from `@site`

Dugyu:
## Motivation

`theme/OgHead.tsx` currently reads shared OG config through `@site`, which behaves as a host-dependent directory alias in cross-repository consumption. That makes it unsuitable as a stable dependency boundary for shared configuration modules.

This PR introduces a dedicated alias for shared OG config so the dependency is explicit, semantic, and independently configurable by different hosts.

### `rspress.config.ts`
- Add a dedicated `@og-config` alias for `shared-og-config.ts`

### `theme/OgHead.tsx`
- Replace `@site/shared-og-config` with `@og-config`

### `tsconfig.json`
- Add the matching TypeScript path mapping for `@og-config`

## Additional notes

- This change only updates import resolution and dependency boundaries.
- It does not change OG image selection logic, canonical URL logic, or generated meta tag behavior.
- `@site` remains unchanged in this PR, but shared config modules no longer depend on it as a general directory alias.

## Validation

- Run `pnpm build`
- Confirm `theme/OgHead.tsx` no longer imports `@site/shared-og-config`
- Verify generated canonical URLs and Open Graph tags remain unchanged

## Related

- #1328 

Change-Id: I088d1b6642343e8fe95f58ad2b3ab50d35f06803

- Add the `@og-config` alias to `rspress.config.ts` and `tsconfig.json`.
- Use `@og-config` to import shared Open Graph configuration in `theme/OgHead.tsx`.

### DIFF
--- a/rspress.config.ts
+++ b/rspress.config.ts
@@ -68,6 +68,7 @@ export default defineConfig({
           'node_modules/react-render-to-markdown/dist/index.js',
         ),
         '@site': path.join(__dirname),
+        '@og-config$': path.join(__dirname, 'shared-og-config.ts'),
         '@': path.join(__dirname, 'src'),
         '@docs': path.join(__dirname, 'sharedDocs', 'packageDocs'),
         '@assets': path.join(__dirname, 'docs', 'public', 'assets'),
--- a/theme/OgHead.tsx
+++ b/theme/OgHead.tsx
@@ -16,12 +16,7 @@ import {
   useLocation,
   usePageData,
 } from '@rspress/core/runtime';
-import {
-  OG_BASE,
-  getCover,
-  ogAbsolute,
-  selectOg,
-} from '@site/shared-og-config';
+import { OG_BASE, getCover, ogAbsolute, selectOg } from '@og-config';
 
 const TWITTER_SITE = '@LynxJS_org';
 /** Keep og:description bounded — blog descriptions can be long. */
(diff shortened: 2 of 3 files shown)


Huxpro (dismissed):
The standalone change is small and CI is green, but this PR changes a **cross-repository consumer contract**, so I don't think it is low-risk to merge yet.

#1328 says the motivating failure happens when the OSS theme is consumed by `lynx-website-inhouse`, where `@site` resolves to the host repo, and its validation checklist explicitly calls out verifying that the in-house repo can configure migrated aliases. This PR changes `OgHead.tsx` to require a new bare specifier, `@og-config`, but defines that alias only in this repository's `rspress.config.ts` / `tsconfig.json`. A host that consumes the theme with its own Rspress config will fail to resolve `@og-config` unless its config lands in lockstep.

Please link/land the corresponding host-side alias change, or add a representative cross-repo consumer fixture/test (or compatibility fallback) that proves the migrated theme resolves outside the standalone website. The current `pnpm build` only proves the OSS host, not the behavior this PR is specifically intended to fix.

Minor follow-up while touching this: because `@og-config` aliases a single file, consider making the bundler alias exact (`@og-config$`) so accidental subpath imports cannot be treated as valid.

Dugyu:
Thanks, addressed in the in-house consumer:

- Added matching exact `@og-config$` Rspack and TypeScript path aliases in `lynx-website-inhouse`.
- Merged the in-house change before this update (internal PR1225)
- Updated the OSS alias to `@og-config$` as well, so only the intended module entry point resolves.

This keeps the OSS and in-house resolver contracts aligned for the next submodule sync.

PR #1333, opened 2026-08-22
fix(docs): make published code snippets compile

Shouqun:
Fixes documentation defects on `release/4.0`.

- **macOS guide does not compile.** Every `@implementation` block is
  unterminated (`missing '@end'`), `applicationDidFinishLaunching` carries a
  literal backslash before the pointer star, `ExampleGenericResourceFetcher`
  names `LynxResourceRequest` / `LynxResourceResponse` in `lynx::pub` rather than
  `lynx::pub::resource`, and the fragment ends with an unmatched opening fence.
- **`viewpager-API` carries mangled JSDoc** in both locales: the leading `* `
  continuation markers were rewritten as `_`, leaving comment blocks that no
  longer parse. The Chinese page keeps its Chinese descriptions.
- **Objective-C escapes** in the custom component guide.
- **Dead links in `AGENTS.md`**.
- **Truncated version** in the 3.2 blog post.

## Verification

- **macOS** — every Objective-C++ block is extracted from the `.mdx` and compiled
  with `clang -fsyntax-only -fobjc-arc` against the released macOS SDK headers.
  As published **none of the five compile**; at this head **all five do**, in both
  locales.
- **Android** — the guide's dependency block is handed to Gradle against
  `google()` + `mavenCentral()` to confirm every documented coordinate resolves.
- **Links** — checked against the real route table, modelling `route.exclude`,
  `.html`/`.mdx` equivalence, client redirects and Netlify 301s, so only genuinely
  dead links are reported.
- **CI locally** — `check-case-collisions`, `check-asset-urls`,
  `check-lynx-ui-routes`, `pnpm install --frozen-lockfile`, `pnpm run format:check`
  and `pnpm run build` all pass on this branch.

## Series

`main` #1332 · `release/4.0` #1333 · `release/3.9` #1334 · `release/3.8` #1346 ·
`release/3.7` #1335 · `release/3.6` #1341 · `release/3.5` #1340 ·
`release/3.4` #1339.

The 3.2 blog line (`lynx-service-image:3.2.0-`) is now corrected on **every**
published branch, including `release/3.8` via #1346, so no site version disagrees
with another.

## Follow-up split out

The element-API `invoke()` example convention — type declarations written where
runnable code is shown — is tracked in #1347 with a measured inventory across all
five branches, the two candidate representations, and a coordinated landing plan.
It is not fixed here because doing so would pull in `text.mdx`, `list.mdx`,
`image.mdx` and `webview-API.mdx` — files untouched by the defects these PRs
[...]

PR #1332, opened 2026-08-22
fix(docs): make published code snippets compile and resolve

Shouqun:
Fixes documentation defects on `main`.

- **macOS guide does not compile.** Every `@implementation` block is
  unterminated (`missing '@end'`), `applicationDidFinishLaunching` carries a
  literal backslash before the pointer star, `ExampleGenericResourceFetcher`
  names `LynxResourceRequest` / `LynxResourceResponse` in `lynx::pub` rather than
  `lynx::pub::resource`, and the fragment ends with an unmatched opening fence.
- **`viewpager-API` carries mangled JSDoc** in both locales.
- **Lynxtron examples import a package that does not exist.** Six examples
  import from `lynxtron`, which is not published; the project scaffolded by
  `create @lynx-js/lynxtron` imports from `@lynx-js/lynxtron`.
- **Objective-C escapes** in the custom component guide.
- **Dead links in `AGENTS.md`**.
- **Truncated version** in the 3.2 blog post.

## Verification

- **macOS** — every Objective-C++ block is extracted from the `.mdx` and compiled
  with `clang -fsyntax-only -fobjc-arc` against the released macOS SDK headers.
  As published **none of the five compile**; at this head **all five do**, in both
  locales.
- **Android** — the guide's dependency block is handed to Gradle against
  `google()` + `mavenCentral()` to confirm every documented coordinate resolves.
- **Links** — checked against the real route table, modelling `route.exclude`,
  `.html`/`.mdx` equivalence, client redirects and Netlify 301s, so only genuinely
  dead links are reported.
- **CI locally** — `check-case-collisions`, `check-asset-urls`,
  `check-lynx-ui-routes`, `pnpm install --frozen-lockfile`, `pnpm run format:check`
  and `pnpm run build` all pass on this branch.

## Series

`main` #1332 · `release/4.0` #1333 · `release/3.9` #1334 · `release/3.8` #1346 ·
`release/3.7` #1335 · `release/3.6` #1341 · `release/3.5` #1340 ·
`release/3.4` #1339.

The 3.2 blog line (`lynx-service-image:3.2.0-`) is now corrected on **every**
published branch, including `release/3.8` via #1346, so no site version disagrees
with another.

## Follow-up split out

The element-API `invoke()` example convention — type declarations written where
runnable code is shown — is tracked in #1347 with a measured inventory across all
five branches, the two candidate representations, and a coordinated landing plan.
It is not fixed here because doing so would pull in `text.mdx`, `list.mdx`,
[...]

PR #1334, opened 2026-08-22
fix(docs): pin PRIMJS_VERSION to a published version and repair code snippets

Shouqun:
Fixes documentation defects on `release/3.9`, including one that makes the
integration guide unbuildable on both mobile platforms.

- **`PRIMJS_VERSION` points at a version that was never published.**
  `docs/public/version.json` pins `3.9.0`, but no PrimJS 3.9.0 exists. Every
  Android and iOS dependency snippet on the branch renders from this value, so
  the guide fails at dependency resolution on both platforms before any code
  runs. Lynx 3.9.0 depends on `primjs 3.8.0-alpha.6` on Android, and the
  `Lynx/Framework` 3.9.0 podspec pins `PrimJS/quickjs` to exactly that version —
  it is the only value that satisfies both Gradle and CocoaPods.
- **macOS guide does not compile** — unterminated `@implementation` blocks, an
  escaped pointer star in `applicationDidFinishLaunching`, the
  `lynx::pub::resource` namespace, and an unmatched trailing fence.
- **Objective-C escapes** in the custom component guide.
- **Dead links in `AGENTS.md`**.
- **Truncated version** in the 3.2 blog post.

With the version fix, the Android app builds and runs against 3.9.0 and the iOS
pods resolve and build; both were checked end to end.

## Verification

- **macOS** — every Objective-C++ block is extracted from the `.mdx` and compiled
  with `clang -fsyntax-only -fobjc-arc` against the released macOS SDK headers.
  As published **none of the five compile**; at this head **all five do**, in both
  locales.
- **Android** — the guide's dependency block is handed to Gradle against
  `google()` + `mavenCentral()` to confirm every documented coordinate resolves.
- **Links** — checked against the real route table, modelling `route.exclude`,
  `.html`/`.mdx` equivalence, client redirects and Netlify 301s, so only genuinely
  dead links are reported.
- **CI locally** — `check-case-collisions`, `check-asset-urls`,
  `check-lynx-ui-routes`, `pnpm install --frozen-lockfile`, `pnpm run format:check`
  and `pnpm run build` all pass on this branch.

## Series

`main` #1332 · `release/4.0` #1333 · `release/3.9` #1334 · `release/3.8` #1346 ·
`release/3.7` #1335 · `release/3.6` #1341 · `release/3.5` #1340 ·
`release/3.4` #1339.

The 3.2 blog line (`lynx-service-image:3.2.0-`) is now corrected on **every**
published branch, including `release/3.8` via #1346, so no site version disagrees
with another.

## Follow-up split out

The element-API `invoke()` example convention — type declarations written where
runnable code is shown — is tracked in #1347 with a measured inventory across all
five branches, the two candidate representations, and a coordinated landing plan.
It is not fixed here because doing so would pull in `text.mdx`, `list.mdx`,
`image.mdx` and `webview-API.mdx` — files untouched by the defects these PRs
exist for — and land a convention change on five release branches without
coordination.

### DIFF
--- a/docs/en/blog/lynx-3-2.mdx
+++ b/docs/en/blog/lynx-3-2.mdx
@@ -258,7 +258,7 @@ dependencies {
     implementation("org.lynxsdk.lynx:primjs:2.12.0")
 
     // integrating image-service
-    implementation("org.lynxsdk.lynx:lynx-service-image:3.2.0-")
+    implementation("org.lynxsdk.lynx:lynx-service-image:3.2.0")
 
     // image-service dependencies, if not added, images cannot be loaded; if the host APP needs to use other image libraries, you can customize the image-service and remove this dependency
     implementation("com.facebook.fresco:fresco:2.3.0")
--- a/docs/en/guide/custom-native-component/custom-component-iOS.mdx
+++ b/docs/en/guide/custom-native-component/custom-component-iOS.mdx
@@ -275,7 +275,7 @@ LYNX_LAZY_REGISTER_UI("explorer-input")
   self.view.padding = self.padding;
   }
 
-LYNX_PROP_SETTER("value", setValue, NSString \*) {
+LYNX_PROP_SETTER("value", setValue, NSString *) {
 self.view.text = value;
 }
 
@@ -295,7 +295,7 @@ self.view.text = value;
   }
 
 - (void)setPadding:(UIEdgeInsets)padding {
-  \_padding = padding;
+  _padding = padding;
   [self setNeedsLayout];
   }
 
--- a/docs/en/guide/start/fragments/macos/integrating-lynx-with-existing-app-macos.mdx
+++ b/docs/en/guide/start/fragments/macos/integrating-lynx-with-existing-app-macos.mdx
@@ -94,6 +94,7 @@ target_link_options(${PROJECT_NAME} PRIVATE
 - Lynx Service needs to be actively injected
 
 <CodeFold height={360} toggle>
+
 ```objective-c title=AppDelegate.mm {17,20-30}
 #import "AppDelegate.h"
 #import "ViewController.h"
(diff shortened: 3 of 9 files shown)


Huxpro (changes requested):
The `PRIMJS_VERSION` correction itself looks well-founded: `PrimJS 3.8.0-alpha.6` is published, and Lynx’s CocoaPods spec pins `PrimJS/quickjs` to that exact version. I’m still keeping this head blocked on two documentation consistency issues:

1. The second macOS `AppDelegate.mm` / LynxEnv block still references `LynxHttpServiceImpl` without defining it. The PR description explicitly says 1 of the 5 extracted Objective-C++ samples still does not compile. Please make that block self-contained or clearly identify it as an incremental continuation of the preceding AppDelegate instead of presenting it as a complete file.

2. The duplicated Lynx 3.2 blog typo is being fixed on main/4.0/3.9/3.7, but the same `lynx-service-image:3.2.0-` line is still present on `release/3.8`, `release/3.6`, `release/3.5`, and `release/3.4`. Please add or track the companion backports so this historical post remains aligned across the versioned website.

CI is green on this exact head; the blockers are content correctness and cross-site consistency.

Huxpro (changes requested):
The current head resolves the earlier macOS sample blocker by making the later `AppDelegate.mm` example self-contained, and the `PRIMJS_VERSION = 3.8.0-alpha.6` correction remains consistent with the 3.9 dependency floor previously verified. CI is green on this exact SHA.

One substantive cross-version blocker remains: the duplicated Lynx 3.2 blog post still contains `lynx-service-image:3.2.0-` on `release/3.8`, and there is still no open 3.8 companion PR. The other published branches are now covered by this series (#1332–#1335 and #1339–#1341), so leaving 3.8 behind would make the same historical post knowingly inconsistent depending on the selected site version. Please land/link the 3.8 backport with the series before merging.

Minor housekeeping: the PR description still says only 4 of 5 extracted macOS samples compile even though the current diff now supplies `LynxHttpServiceImpl` to the previously failing later block; please refresh that verification text so it describes the actual head.

Huxpro (approved):

PR #1335, opened 2026-08-22
fix(docs): repair mangled code blocks and dead links

Shouqun:
Fixes documentation defects on `release/3.7` that stop published snippets from
working when copied.

- **macOS guide does not compile.** Every `@implementation` block is
  unterminated (`missing '@end'`), `applicationDidFinishLaunching` carries a
  literal backslash before the pointer star so `NSNotification` parses as passed
  by value, `ExampleGenericResourceFetcher` names `LynxResourceRequest` /
  `LynxResourceResponse` in `lynx::pub` instead of `lynx::pub::resource`, and the
  fragment ends with an unmatched opening fence.
- **Element API pages carry mangled JSDoc.** In `input-API`, `textarea-API` and
  `scroll-view-API` the leading `* ` continuation markers were rewritten as `_`,
  so the comment blocks no longer parse. The replacements are the
  already-correct blocks from `release/3.8`, which are semantically identical
  here.
- **Objective-C escapes** leak into the custom component guide (`NSString \*`,
  `\_padding`). Backslash escapes are not processed inside code fences, so these
  reach the reader verbatim.
- **Dead links in `AGENTS.md`** — this file is served to coding agents, so each
  stale path is a 404 for every agent that follows it.
- **Truncated version** in the 3.2 blog post (`lynx-service-image:3.2.0-`).

## Verification

- **macOS** — every Objective-C++ block is extracted from the `.mdx` and compiled
  with `clang -fsyntax-only -fobjc-arc` against the released macOS SDK headers.
  As published **none of the five compile**; at this head **all five do**, in both
  locales.
- **Android** — the guide's dependency block is handed to Gradle against
  `google()` + `mavenCentral()` to confirm every documented coordinate resolves.
- **Links** — checked against the real route table, modelling `route.exclude`,
  `.html`/`.mdx` equivalence, client redirects and Netlify 301s, so only genuinely
  dead links are reported.
- **CI locally** — `check-case-collisions`, `check-asset-urls`,
  `check-lynx-ui-routes`, `pnpm install --frozen-lockfile`, `pnpm run format:check`
  and `pnpm run build` all pass on this branch.

## Series

`main` #1332 · `release/4.0` #1333 · `release/3.9` #1334 · `release/3.8` #1346 ·
`release/3.7` #1335 · `release/3.6` #1341 · `release/3.5` #1340 ·
`release/3.4` #1339.

The 3.2 blog line (`lynx-service-image:3.2.0-`) is now corrected on **every**
published branch, including `release/3.8` via #1346, so no site version disagrees
with another.

## Follow-up split out

The element-API `invoke()` example convention — type declarations written where
runnable code is shown — is tracked in #1347 with a measured inventory across all
five branches, the two candidate representations, and a coordinated landing plan.
It is not fixed here because doing so would pull in `text.mdx`, `list.mdx`,
`image.mdx` and `webview-API.mdx` — files untouched by the defects these PRs
exist for — and land a convention change on five release branches without
coordination.

### DIFF
--- a/docs/en/api/elements/built-in/input-API.mdx
+++ b/docs/en/api/elements/built-in/input-API.mdx
@@ -295,50 +295,50 @@ Require focus
 ```ts
 
 lynx.createSelectorQuery()
-.select('#id')
-.invoke({
-method: 'getValue',
-success: Callback<{
-/**
-_ Is composing or not, iOS only
-_ @Android
-_ @iOS
-_ @Harmony
-_ @Web
-_ @since 3.4
-\*/
-isComposing: boolean;
-/**
-_ End position of the selection
-_ @Android
-_ @iOS
-_ @Harmony
-_ @Web
-_ @since 3.4
-_/
-selectionEnd: number;
-/\*\*
-_ Begin position of the selection
-_ @Android
-_ @iOS
-_ @Harmony
-_ @Web
-_ @since 3.4
-_/
-selectionStart: number;
-/\*\*
-_ Input content
-_ @Android
-_ @iOS
-_ @Harmony
-_ @Web
-_ @since 3.4
-\*/
-value: string;
-}>;
-fail: function (res) {},
-})
-.exec();
+     .select('#id')
+     .invoke({
+      method: 'getValue',
+      success: Callback<{
+        /**
+         * Is composing or not, iOS only
+         * @Android
+         * @iOS
+         * @Harmony
+         * @Web
+         * @since 3.4
+         */
+        isComposing: boolean;
+        /**
+         * End position of the selection
+         * @Android
+         * @iOS
+         * @Harmony
+         * @Web
+         * @since 3.4
+         */
+        selectionEnd: number;
+        /**
+         * Begin position of the selection
+         * @Android
+         * @iOS
+         * @Harmony
+         * @Web
+         * @since 3.4
+         */
+        selectionStart: number;
+        /**
+         * Input content
+         * @Android
+         * @iOS
+         * @Harmony
+         * @Web
+         * @since 3.4
+         */
+        value: string;
+      }>;
+      fail: function (res) {},
+    })
+    .exec();
 
 ````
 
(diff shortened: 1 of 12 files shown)


Huxpro (changes requested):
I’m keeping this release/3.7 repair blocked on three correctness/consistency points:

1. The touched `input-API`, `textarea-API`, and `scroll-view-API` blocks now have repaired JSDoc markers, but they are still embedded as executable `lynx.createSelectorQuery().invoke(...)` snippets while containing type-only syntax such as `success: Callback<{...}>;`, `value: string;`, `rate: number;`, and optional-property declarations. If these blocks are intended to be copyable invocation examples, please use actual runtime values/callbacks; if they are intended to describe the API shape, separate that type declaration from the invocation example. Merely copying the release/3.8 formatting preserves this ambiguity.

2. The second macOS `AppDelegate.mm` / LynxEnv block still uses `LynxHttpServiceImpl` without declaring it. The PR description itself acknowledges that one extracted sample still fails to compile. Please make it self-contained or explicitly label it as an incremental continuation of the preceding AppDelegate.

3. The Lynx 3.2 blog typo is still present on `release/3.8`, `release/3.6`, `release/3.5`, and `release/3.4`, while this series covers main/4.0/3.9/3.7. Please add or track companion backports so the duplicated historical content stays aligned across versioned sites.

I also do not see a normal GitHub Actions PR workflow run attached to this exact `2ba58dad` head yet, so I would want the standard branch validation in addition to the content fixes before merging.

Huxpro (changes requested):
The current head fixes the earlier macOS self-containment issue, but I’m still keeping this release/3.7 repair blocked on executable-example correctness and cross-version consistency.

1. The repaired `input`, `textarea`, and `scroll-view` blocks are still not valid invocation examples. They now have clean JSDoc formatting, but the code passed to `.invoke(...)` still contains type-only constructs such as `success: Callback<{ ... }>;`, `value: string;`, `rate: number;`, and `offset?: number;`, plus semicolon-terminated `params` objects. Those declarations cannot execute as JavaScript/TypeScript object values. Please either make these real runtime examples with concrete values/callback functions, or clearly separate the parameter/result type declarations from the executable invocation snippets. Apply the same representation consistently in EN/ZH.

2. The duplicated Lynx 3.2 post is still broken on `release/3.8`, and there is still no open companion PR for that branch. Please land/link the 3.8 backport with the rest of the series so the historical post stays aligned across every published version.

I also do not see a normal PR CI workflow run attached to this exact SHA yet. Even after the content issues are fixed, please make sure the usual validation runs on the final head.

Huxpro (approved):

PR #1336, opened 2026-08-22
docs(compat-data): add WebAssembly compat data

orionmiz:
# Summary

WebAssembly support currently exists in the docs only as a `<VersionBadge v={3.8} />` and prose on the [WebAssembly guide page](https://lynxjs.org/guide/scripting-runtime/webassembly.html). This PR turns it into structured Lynx Compatibility Data so it can be rendered through `APISummary`/`APITable` and picked up by downstream consumers of `@lynx-js/lynx-compat-data`.

## Changes

**`docs(compat-data): add WebAssembly compat data`**

- Adds `packages/lynx-compat-data/lynx-api/global/WebAssembly.json` with a single `lynx-api.global.WebAssembly` entry (`lynx_path: guide/scripting-runtime/webassembly`).
- Regenerates `api-stats.json` via `pnpm run gen-stats` (the large diff is mostly feature-id renumbering, same as previous data PRs).

**`docs(guide): render WebAssembly compat table from LCD`**

- Adds a "Compatibility" section with `<APITable query="lynx-api/global/WebAssembly" />` as the final section of the en/zh WebAssembly guide pages, matching the heading and placement used by other guide pages (devtool, trace, recorder).

## Data and sources

Every statement is sourced from documentation in this repository; nothing is inferred:

| Platform | Value | Basis |
| --- | --- | --- |
| `android` | `"3.8"` + `partial_implementation: true` | [guide/scripting-runtime/webassembly](https://lynxjs.org/guide/scripting-runtime/webassembly.html): basic support since Lynx 3.8 via PrimJS, background thread only, and "the complete WebAssembly JavaScript API is not available yet" (no `compile`/`instantiate`/`validate`/streaming variants) — captured in `notes` per the schema requirement for partial implementations. |
| `ios` | `null` (unknown) | The 3.8 integration is Android-only, but that alone does not prove the `WebAssembly` global is absent from Lynx's default iOS background JavaScriptCore context (JSC ships its own WebAssembly implementation), so iOS stays unknown until verified against the runtime. |
| `harmony`, `clay_macos`, `clay_windows` | `null` (unknown) | No documented basis in this repository. |
| `web_lynx` | `null` (unknown) | The background thread on Web runs in a real browser worker, so the native `WebAssembly` global is plausibly reachable — but I could not find a documented or audited basis in this repo (the web_lynx audit trail in #1279 covers other APIs), so I left it unknown rather than claiming support. Happy to flip it to `true` with a note if a maintainer can confirm the web-core sandbox exposes it. |

`status.experimental: true` follows the schema guidance for recently shipped features (3.8, 2026-05).

## Verification

- `pnpm --filter @lynx-js/lynx-compat-data run lint` (schema validation) ✅
- `pnpm --filter @lynx-js/lynx-compat-data run lint:keys` ✅
- `pnpm --filter @lynx-js/lynx-compat-data run test` ✅
- `pnpm run format:check` ✅
- `pnpm run build` ✅

### DIFF
--- a/docs/en/guide/scripting-runtime/webassembly.mdx
+++ b/docs/en/guide/scripting-runtime/webassembly.mdx
@@ -2,7 +2,7 @@
 description: 'WebAssembly support in Lynx via the PrimJS runtime on Android.'
 ---
 
-import { Go, VersionBadge } from '@lynx';
+import { APITable, Go, VersionBadge } from '@lynx';
 
 # WebAssembly
 
@@ -54,3 +54,7 @@ Try the complete Lynx example below to see the same flow in a page:
 />
 
 For more details about the underlying runtime support, see the [PrimJS WebAssembly document](https://github.com/lynx-family/primjs/blob/develop/docs/wasm.md).
+
+## Compatibility
+
+<APITable query="lynx-api/global/WebAssembly" />
--- a/docs/zh/guide/scripting-runtime/webassembly.mdx
+++ b/docs/zh/guide/scripting-runtime/webassembly.mdx
@@ -2,7 +2,7 @@
 description: 'Lynx 3.8 起通过 PrimJS 支持 WebAssembly。'
 ---
 
-import { Go, VersionBadge } from '@lynx';
+import { APITable, Go, VersionBadge } from '@lynx';
 
 # WebAssembly
 
@@ -54,3 +54,7 @@ const result = instance.exports.add(1, 2);
 />
 
 更多底层运行时支持的说明，可参考 [PrimJS 的 WebAssembly 文档](https://github.com/lynx-family/primjs/blob/develop/docs/wasm.md)。
+
+## 兼容性
+
+<APITable query="lynx-api/global/WebAssembly" />
--- a/packages/lynx-compat-data/api-stats.json
+++ b/packages/lynx-compat-data/api-stats.json
@@ -172,7 +172,7 @@
           "clay": 100
         },
         "exclusive": {
-          "android": 0,
+          "android": 1,
           "ios": 0,
           "harmony": 0,
           "web_lynx": 0,
@@ -620,7 +620,7 @@
       "android": {
         "supported_count": 1010,
         "coverage_percent": 89,
-        "exclusive_count": 18
+        "exclusive_count": 19
       },
       "ios": {
         "supported_count": 1011,
@@ -39625,7 +39625,7 @@
           "clay": 100
         },
         "exclusive": {
-          "android": 0,
+          "android": 1,
           "ios": 0,
           "harmony": 0,
           "web_lynx": 0,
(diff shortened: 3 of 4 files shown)


Huxpro (changes requested):
Thanks for turning this into structured compatibility data. I’m blocking this head on one compatibility source-of-truth issue.

`ios.version_added: false` is stronger than the evidence cited here. In LCD, `false` means the feature is definitively unsupported, while `null` means support is unknown. The existing Lynx docs establish that the **Lynx 3.8 PrimJS WebAssembly integration** was added on Android, and upstream lynx#5294 is likewise explicitly Android-only (Android/JNI + `primjsWasm`). That proves the Android integration/floor, but it does not by itself prove that the `WebAssembly` global is absent from Lynx’s default iOS background JavaScriptCore context. In fact, JavaScriptCore itself has WebAssembly support, so this needs a Lynx/iOS-specific runtime basis rather than an inference from the Android-only integration.

Please either (a) provide implementation/test evidence from a Lynx iOS background `JSContext` showing `globalThis.WebAssembly` is unavailable and cite that basis in the note, or (b) keep iOS at `version_added: null` until it is verified. The Android `3.8` + partial-implementation statement otherwise matches the current guide and implementation direction well.

Separately, the normal CI and Compat Data Stats runs on this exact head currently ended as `action_required`, and the CLA check is still pending, so I would not merge this head yet even after the metadata issue is resolved.

orionmiz:
@Huxpro You're right that `false` was overclaiming. Went with (b): iOS is `null` as of 2a0ab80, and the note implying absence is gone.

CLA is signed now. The CI and stats runs on this head still need approval on your side.

One thing I considered but skipped: splitting the statics/interfaces into subfeatures like BCD's `webassembly/api`. The android column would be sourced from the guide, but every other platform would just be a stack of unknowns right now, so I kept the single entry. Can do that as a follow-up if you think it's worth having.

Huxpro (approved):
The compatibility source-of-truth issue from the previous review is resolved on this head: iOS is now correctly left as `null`/unknown rather than being asserted unsupported, while Android retains the documented Lynx 3.8 partial/background-thread-only boundary. EN/ZH remain aligned and the structured table matches the existing badge/prose. I also resolved the CodeRabbit thread asking for `ios: false`, because the current PR explicitly does not claim iOS is unsupported and there is still no Lynx/iOS-specific evidence for that stronger value.

Approving the documentation/content on this exact head. I am not merging yet because the normal CI and Compat Data Stats runs for `2a0ab802` still ended as `action_required` instead of executing; please get the standard workflows to run before merge.

Huxpro:
Merged via #1344. The replacement branch pointed at this PR's exact head commit (`2a0ab8024f9ac54a498d12a4dc7835b809328461`), preserving all three commits and their authorship; it was needed because the fork's required Actions runs remained in `action_required` and could not be approved by the repository GitHub App.

PR #1337, opened 2026-08-24
fix(compat-data): repoint stale lynx_path values after docs restructuring

orionmiz:
## Motivation

`lynx_path` is the "Docs" link the compat table renders (`withBase('/' + lynxPath)`) and the source of `doc_url` in `api-stats.json`. The docs restructuring moved and renamed pages without updating it, so those links 404.

Checking every `lynx_path` in the git-tracked compat data against the routes `docs/en/**` publishes found **43 dead entries across 29 files**.

Removing `lynx_path` from an undocumented API did not help, because `generate-stats.ts` fell back to `generateDocUrl()` and synthesised the same nonexistent route. That fallback was already emitting dead URLs for features that never had a `lynx_path` — `/api/elements/built-in/x-foldview-ng`, `/api/lynx-api/fetch/Headers`, every `layout-animation-*` property, `/guide/devtool/logbox`, and others.

Targets were derived from the repo, not guessed: 441 pages declare `api: <compat key>` in their frontmatter, which is authoritative for that key. Anchors reuse the slug scheme already present in in-repo links (`#elementsetstyleproperty`, `#entrytype`). Where neither could confirm a target, the link points at the page rather than an invented fragment.

### `packages/lynx-compat-data/**/*.json`

- Repoint 38 stale `lynx_path` values at the moved page
- Drop `lynx_path` from 5 entries no page documents (`lynx.{get,set}SharedData`, `lynx.{register,remove}SharedDataObserver`, `lynx-view.update-screen-metrics`)
- Remove the `Animation.finish()` entry — not a real API, see below

<details><summary>Full before → after list</summary>

| before | after |
| --- | --- |
| `api/lynx-api/intersection-observer/intersection-observer` | `api/lynx-api/intersection-observer` |
| `api/lynx-api/intersection-observer/disconnect` | `…/intersection-observer-disconnect` |
| `api/lynx-api/intersection-observer/observe` | `…/intersection-observer-observe` |
| `api/lynx-api/intersection-observer/relative-to` | `…/intersection-observer-relative-to` |
| `api/lynx-api/intersection-observer/relative-to-screen` | `…/intersection-observer-relative-to-screen` |
| `api/lynx-api/intersection-observer/relative-to-viewport` | `…/intersection-observer-relative-to-viewport` |
| `api/lynx-api/lynx/lynx-animate_api` | `api/lynx-api/lynx/lynx-animate-api` |
| `api/lynx-api/lynx/queueMicrotask` | `api/lynx-api/lynx/lynx-queue-microtask` |
| `api/lynx-api/lynx/resume-exposure` | `api/lynx-api/lynx/lynx-resume-exposure` |
| `api/lynx-api/lynx/stop-exposure` | `api/lynx-api/lynx/lynx-stop-exposure` |
| `api/lynx-api/lynx/set-observer-frame-rate` | `api/lynx-api/lynx/lynx-set-observer-frame-rate` |
| `api/lynx-api/main-thread/lynx-query-selector-all` | `api/lynx-api/main-thread/lynx-query-select-all` |
| `api/lynx-api/main-thread/main-thread` | `api/lynx-api/main-thread` |
| `api/lynx-api/main-thread/ElementGetComputedStyle` | `api/lynx-api/main-thread/element-get-computed-style` |
| `api/lynx-api/main-thread/ElementAnimate{,Play,Pause,Cancel}` | `api/lynx-api/main-thread/lynx-animate-api` |
| `api/lynx-api/main-thread/element-animate` | `api/lynx-api/main-thread/main-thread-element#elementanimate` |
| `api/lynx-api/nodes-ref/nodes-ref` | `api/lynx-api/nodes-ref` |
| `api/lynx-api/selector-query/selector-query` | `api/lynx-api/selector-query` |
| `api/lynx-api/performance-api/performance-entry/entry-type` | `…/performance-entry#entrytype` |
| `api/lynx-api/performance-api/performance-entry/name` | `…/performance-entry#name` |
| `api/css/properties/custom-properties` | `api/css/properties/css-variable` |
| `guide/networking` | `guide/interaction/networking` |
| `guide/performance/timing-flag` | `guide/performance/monitor-performance/timing-flag` |
| `api/react/main-thread` | `react/main-thread-script` / `api/react/Function.runOnBackground` |

</details>

### `packages/lynx-compat-data/scripts/lib/doc-routes.ts` (new)

- Build the set of routes the site publishes from `docs/en/**`, applying the `route.exclude` patterns from `rspress.config.ts`
- Return `null` when docs sources are absent, so the package stays usable standalone

### `packages/lynx-compat-data/scripts/generate-stats.ts`

- Resolve `doc_url` against that route set instead of emitting it unchecked
- Omit an unresolvable URL — consumers can then tell "undocumented" from "documented over there" instead of following a 404
- Report an unresolvable authored `lynx_path` on stderr; drop an unresolvable generated fallback quietly
- `doc_url` was already optional in `APIInfo`/`RecentAPI` and `src/components/api-status/types.ts`, and nothing reads it yet, so no consumer changes were needed

### `packages/lynx-compat-data/api-stats.json`

- Regenerated

## Review follow-ups

**Dead links in the generated consumer surface.** Fixed as a class, not per instance: dead `doc_url` values in `api-stats.json` go from **100 on `main` to 0**. Verified that nothing valid was lost by diffing every API path's `doc_url` against the base file — 257 dropped, all dead; 0 valid. The 36 that still do not resolve all come from `@lynx-js/css-defines` and now surface as a build warning.

**`Animation.finish()`.** I went looking for the semantics to document and found the method is not part of the API, which is why no page documents it:

- `@lynx-js/types@3.9.0` `types/main-thread/animation.d.ts` declares the MTS `Animation` interface with exactly `cancel()`, `pause()`, `play()`; the background-thread declaration matches
- lynx-stack's worklet runtime `Animation` class has implemented those same three since [`feat: add animate in MTS` (#1534)](https://github.com/lynx-family/lynx-stack/pull/1534) introduced the API, and its `AnimationOperation` enum stops at `CANCEL`
- `AnimationOperation.FINISH` exists only in the element PAPI contract (web-core `constants.ts`, testing-environment mock); nothing reachable from user code sends it

So the EN/ZH pages listing three methods are correct and the compat entry was the inconsistency — removed rather than documenting an uncallable method. **If it tracks a native surface not visible from OSS sources, say so and I will restore it with a `version_added: false` shape.** Dataset goes 1227 → 1226; `compare-stats.mjs` renders this as `Lynx Main Thread API 17 (-1)`.

## Open questions

1. **`react.main-thread.*` targets are inferred.** No `api:` binding exists for these two keys; `react/main-thread-script` and `api/react/Function.runOnBackground` are the only pages covering them. Confirm, or tell me to drop `lynx_path` instead.
2. **`gen-stats` now reads `docs/en/**`.** `AGENTS.md` notes the downstream applies internal overlays after copying OSS content, so the two sites' docs differ and the same compat data can yield different `doc_url` omissions. I think verifying against the routes each site actually publishes is right, but it is a new coupling — your call.
3. **`update-screen-metrics`** could point at `api/lynx-native-api/lynx-view/update-viewport` (which mentions it in passing) if you prefer a near-miss over no link.
4. **`custom-property` and `css-variable` now share a target page.** The page covers both; consolidating the two compat entries is out of scope here.

## Out of scope

`css/properties/` is generated at build time from `@lynx-js/css-defines` and is not tracked in git, so 36 stale paths there cannot be fixed in this repository.

<details><summary>The 36, and why they are two separate defects</summary>

**31 carry a literal `docs/zh/` prefix** — wrong root and wrong locale. Each has a valid en counterpart, so the upstream fix is nearly mechanical (strip the prefix):

`left`, `max-height`, `max-width`, `min-height`, `min-width`, `order`, `padding`, `padding-bottom`, `padding-inline-end`, `padding-inline-start`, `padding-left`, `padding-right`, `padding-top`, `relative-align-bottom`, `relative-align-inline-end`, `relative-align-inline-start`, `relative-align-left`, `relative-align-right`, `relative-align-top`, `relative-bottom-of`, `relative-center`, `relative-id`, `relative-inline-end-of`, `relative-inline-start-of`, `relative-layout-once`, `relative-left-of`, `relative-right-of`, `relative-top-of`, `right`, `top` → `api/css/properties/<name>`

plus `docs/zh/api/css/css-style/position` → `api/css/properties/position`.

**5 have no page at all**: `-x-animation-color-interpolation`, `-x-text-decoration-gap`, `-x-text-decoration-width`, `linear-orientation`, `text-decoration-thickness`.

</details>

Follow-up worth doing separately: `doc-routes.ts` now gives `validate.ts` what it needs to *fail* on a stale `lynx_path` rather than silently degrading it to "no doc link". That would have caught all 43 entries here plus the 31 malformed `css-defines` paths, in the PR that moved the page.

## Validation

- `lint` (schema), `lint:keys`, `test` (10/10), `prettier --check`, `tsc --noEmit --strict` on the changed scripts — all pass
- `compare-stats.mjs`, which the Compat Data Stats workflow runs on every compat-data PR, exits 0 against `main`'s `api-stats.json` and renders the removed API correctly
- Post-change route re-check: 0 dead `lynx_path` in tracked data, 0 dead `doc_url` in `api-stats.json`
- Compat data diff contains no non-`lynx_path` lines apart from the `Animation.finish()` removal

- Repoint stale `lynx_path` values and remove undocumented compatibility entries.
- Validate `doc_url` values against documentation routes and omit unresolved URLs.
- Regenerate `api-stats.json` with zero dead `doc_url` values.
- Support `--docs-root` in `gen-stats` with `LYNX_COMPAT_DOCS_ROOT` fallback.
- Handle missing documentation sources without generating routes.
- Document stats generation and run it during install, development, and build hooks.
- Improve CI stats comparison and upload PR statistics artifacts.


Huxpro (dismissed):
Thanks for the thorough audit — most of the path repoints look like exactly the right cleanup, but I don't think this head is safe to merge yet because it deliberately leaves two source-of-truth inconsistencies behind.

1. Removing `lynx_path` for the four SharedData APIs and `update-screen-metrics` does not actually remove their dead links from the generated consumer surface. `generate-stats.ts` unconditionally falls back to `generateDocUrl(...)` when `compat.lynx_path` is absent, and this PR's regenerated `api-stats.json` therefore still emits `/api/lynx-api/lynx/getSharedData`, `/registerSharedDataObserver`, `/removeSharedDataObserver`, `/setSharedData`, and `/api/lynx-native-api/lynx-view/update-screen-metrics` — the same nonexistent routes this PR is trying to eliminate. Please fix the generator/data model so undocumented APIs can have no `doc_url` (or point them at a real canonical page), and regenerate the stats in the same change. Leaving the tracked compat data "fixed" while the published stats still advertise dead URLs is exactly the kind of cross-site drift we should avoid.

2. `lynx-api/main-thread/ElementAnimate.Animation.finish()` is repointed to `lynx-animate-api`, but that canonical page currently documents only `cancel()`, `pause()`, and `play()` in the returned `Animation` surface; `finish()` is absent. Since the compat entry now intentionally sends readers there, please either document `finish()` in the EN/ZH API reference (with the correct compatibility/version semantics) or do not present that page as adequate documentation for the feature.

The PR description already calls out both gaps, so this is mainly asking that they be resolved in the same consistency repair rather than left as known broken downstream/reference surfaces.

Huxpro (dismissed):
The two blockers from my previous review are addressed on this head: undocumented APIs no longer keep synthesized dead `doc_url`s, and removing `Animation.finish()` is consistent with the current public/runtime `Animation` surface. I found two new issues in the route-validation implementation that still make this unsafe to land as-is.

1. `doc-routes.ts` is platform-dependent. `walk()` feeds `path.relative(base, entryPath)` straight into `toRoute()`, but `toRoute()` and `isExcluded()` only recognize `/`. On Windows, `path.relative()` returns `api\\...` paths, so the route set contains backslashes while every `lynx_path`/generated URL uses `/`; those lookups will fail and `gen-stats` will omit valid `doc_url`s (and warn for valid authored `lynx_path`s). Because `gen-stats` runs from the root `prepare` script, this would make the tracked/generated dashboard data depend on the developer/CI OS. Please normalize filesystem-relative paths to POSIX separators before applying route logic, and add a Windows-style path case to the route tests.

2. This also creates a new downstream contract that has not been validated. `AGENTS.md` explicitly lists `packages/lynx-compat-data/` as a downstream generated-data input, notes that the downstream site applies its own docs overlays / maintains its own Rspress config, and requires downstream validation for changes affecting that contract. The new helper hard-codes both `docs/en` relative to the package's source-tree location and a second copy of `route.exclude` semantics. Depending on whether the downstream executes this package from the OSS source tree, a `file:`/symlinked install, or `node_modules`, it can validate against OSS routes, a different route set, or silently fall back to unchecked URLs. Please make the docs-root/route-exclusion source explicit/shared rather than dependent on package realpath, and validate the updated OSS pin in the downstream consumer before landing.

The path repoints themselves look good; these blockers are specifically about making the new `doc_url` validation deterministic and keeping the OSS/downstream generated-data contract aligned.

Huxpro (dismissed):
The new head fixes the two implementation issues from my previous review: route normalization is now deterministic across Windows/POSIX, and docs-root selection is explicitly caller-provided rather than inferred from the package location or a duplicated `route.exclude` copy. The targeted route tests also cover Windows-style separators.

One cross-site blocker remains before this is safe to land: this PR changes the generated-data contract for `packages/lynx-compat-data`, and the repository's own downstream contract requires validation against the updated OSS pin in the in-house consumer. I don't see that validation linked here yet. Please run/link the downstream prepare/build (or the corresponding in-house PR) and confirm how that consumer invokes `gen-stats`—in particular, that it either passes its own `--docs-root`/`LYNX_COMPAT_DOCS_ROOT` when regenerating stats or intentionally consumes the OSS-generated `api-stats.json` semantics. That is the remaining piece needed to make sure the new omission behavior does not diverge between the public and downstream sites.

Separately, the exact head's normal CI and Compat Data Stats workflows are still `action_required`, so they have not provided the usual validation signal yet. The path corrections themselves continue to look sound.

Huxpro:
I checked the downstream compatibility implications of this head. The updated contract is compatible with the current consumer model, and I’ll follow this merge with an in-house MR that updates the pinned OSS revision and runs the downstream prepare/build validation.

Thank you @orionmiz for the exceptionally thorough audit, the careful follow-ups across all three review rounds, and the clear documentation of the remaining generated CSS paths. This is a very solid consistency cleanup.

Huxpro (dismissed):
The implementation fixes from the last round are in place, and both CI + Compat Data Stats are now green on this exact head. I still don't think this is safe to merge yet because the new generated-data contract is not consistently documented or validated across consumers.

1. `src/components/api-status/README.md` still tells maintainers to run `pnpm --filter @lynx-js/lynx-compat-data run gen-stats` directly. On this head, that path omits `--docs-root`, and `AGENTS.md` explicitly says that omitting a docs root skips route verification and emits URLs unchecked. That means following the dashboard's own maintenance instructions can regenerate the very dead `doc_url` values this PR is intended to eliminate. Please update both manual-generation examples to use `pnpm gen:compat-stats` (or otherwise pass the docs root) so the documented workflow matches the new source-of-truth contract.

2. `AGENTS.md` currently says `gen-stats` “emits a doc_url per API, and only emits one that resolves to a real page,” then a few lines later says that omitting the docs root emits every URL unchecked. Please make the guarantee explicitly conditional on supplying `--docs-root` / `LYNX_COMPAT_DOCS_ROOT` so downstream maintainers are not given contradictory contract text.

3. The downstream compatibility check is still only promised after this merge. The same `AGENTS.md` contract says changes affecting these consumed/generated-data interfaces should be validated against the updated OSS pin in the in-house repository. A post-merge follow-up does not establish that the downstream prepare/build actually accepts this new `gen-stats` behavior before OSS lands it. Please link the downstream validation (or the corresponding in-house change) first, including whether that consumer passes its own docs root or intentionally consumes the OSS-generated `api-stats.json`.

The path repoints and generator behavior themselves look sound now; these remaining blockers are about keeping the maintenance instructions and downstream contract aligned with the implementation.

Automated OpenAI-assisted review note: for a human second opinion, please reach out to me directly/in person.

Huxpro (dismissed):
The latest head fixes the two documentation-contract inconsistencies from my previous review: the API-status README now uses `pnpm gen:compat-stats`, and `AGENTS.md` now makes the route-verification guarantee explicitly conditional on supplying a docs root. The route-validation implementation itself still looks sound.

I’m keeping this blocked on the one remaining cross-site contract issue plus final validation:

1. This PR changes the generated-data contract consumed by the in-house site, and the repository’s own `AGENTS.md` explicitly requires validation against the updated OSS pin when that contract changes. I still do not see linked downstream prepare/build validation or a corresponding in-house change confirming whether that consumer passes its own `--docs-root` / `LYNX_COMPAT_DOCS_ROOT` or intentionally consumes the OSS-generated `api-stats.json`. Please establish that before landing; otherwise `doc_url` omission behavior can diverge between the public and downstream sites.

2. The normal `CI` and `Compat Data Stats` runs for this exact `69fba907` head completed as `action_required` rather than executing, so the final documentation-only follow-up has not received the usual workflow validation signal. Please get the standard checks to run on the final SHA.

Once the downstream contract is demonstrated and the final workflows run, I do not currently see another substantive blocker in the patch.

Huxpro (dismissed):
The latest `f8929a4` head keeps the route-validation implementation and maintenance docs in the improved state from the previous round, but it does not resolve the remaining cross-site contract blocker.

`packages/lynx-compat-data` is an explicit downstream generated-data input, and this PR changes when `doc_url` is omitted based on the caller-supplied docs root. I still do not see linked validation from the in-house consumer confirming that its prepare/build succeeds against this OSS pin and, specifically, whether it passes its own `--docs-root` / `LYNX_COMPAT_DOCS_ROOT` or intentionally consumes the OSS-generated `api-stats.json`. That is material here because the downstream site applies overlays and can publish a different route set; landing without validating the consumer can make the public and downstream `doc_url` surfaces diverge.

The exact `f8929a4` SHA also still lacks the normal validation signal: both `CI` and `Compat Data Stats` completed as `action_required` rather than running. Netlify is green, but for a generated compatibility-data contract change I want the standard workflows plus the downstream consumer check before merge.

Once those two validations are demonstrated, I do not currently see another substantive blocker in the patch itself.

This is a Codex-assisted automated review. For a human second opinion, please reach out to me in person.

Huxpro (commented):
Update on the latest head: both the normal CI and Compat Data Stats workflows are now green on `9cc7bc1`, so the earlier CI-validation concern is resolved.

The existing Request Changes still applies only to the remaining explicit downstream-contract requirement: this PR changes the generated `doc_url` behavior consumed by the in-house site, and the repository contract calls for validating the updated OSS pin in that downstream consumer before landing. I still do not see linked downstream prepare/build validation or an in-house change confirming whether that consumer supplies its own docs root or intentionally consumes the OSS-generated semantics.

No new patch-level defect found on this head.

This is a Codex-assisted review. For a human second opinion, please contact me in person.

Huxpro (commented):
Maintainer decision needed

The concrete patch defects from earlier review rounds are fixed, the maintenance docs now match the new docs-root behavior, and both CI + Compat Data Stats are green on this exact head. I also see the maintainer note that the downstream compatibility implications were checked and that an in-house MR/build will follow after OSS lands.

The remaining risk is timing, not a demonstrated defect: `packages/lynx-compat-data` is a downstream generated-data input, and `AGENTS.md` asks for validation against the updated OSS pin when this contract changes. No linked downstream prepare/build has been shown yet, so there is still a chance the in-house consumer exercises `gen-stats` differently and produces different `doc_url` omission behavior.

Options: (1) merge now based on the green OSS checks plus the maintainer's compatibility assessment, then validate the pinned in-house consumer as planned; or (2) hold until that downstream prepare/build is linked. My recommendation is (2) if the downstream validation can be completed promptly, because it closes the only remaining uncertainty; otherwise I would not keep this PR blocked on it absent a concrete incompatibility.

This is a Codex-assisted review. For a human second opinion, please contact me in person.

Dugyu (changes requested):
Further lynx-compat-data-related changes in lynx-website OSS are temporarily blocked. The inhouse sync to `1573bc7` is failing during build (lynx-compat-data related) and is under investigation.

Dugyu (approved):
Approved. The inhouse adaptation for the new stats generator contract is ready: it provides the `ts-node` ESM runner and passes the final `docs/en` as `--docs-root`. After merge, we will update the submodule and validate the full inhouse prepare/build flow against the merged revision.

PR #1350, opened 2026-08-25

### LINKED ISSUE #1347: Element API invoke() examples mix type declarations into runnable code
Follow-up split out of the review on #1332 / #1333 / #1335, which asked that the
element-API example convention be normalized consistently rather than one page
at a time.

## What the problem is

Element API pages document `invoke()` calls with blocks that mix a *type shape*
into what reads as *runnable code*:

```ts
lynx.createSelectorQuery()
  .select('#id')
  .invoke({
    method: 'setValue',
    params: {
      /** Input content */
    };                    // terminates the property with `;` instead of `,`
    success: function (res) {},
  })
  .exec();
```

The block is presented as something you can paste, and it does not parse.

## Measured scope

Counted with `esbuild`'s TypeScript parser over every fenced block containing
`createSelectorQuery` / `.invoke(`:

| Branch | en checked / failing | zh checked / failing |
| ------ | -------------------- | -------------------- |
| `main` | 43 / **20** | 52 / **28** |
| `release/4.0` | 39 / **20** | 48 / **28** |
| `release/3.9` | 31 / **15** | 40 / **23** |
| `release/3.8` | 31 / **15** | 40 / **23** |
| `release/3.7` | 29 / **15** | 20 / **13** |

Failure causes on `main`, both locales:

| Cause | Count | Meaning |
| ----- | ----- | ------- |
| `Expected "}" but found ";"` | 20 | `name: type;` inside `params` |
| `Expected "}" but found "?"` | 10 | optional property declarations, e.g. `name?: string;` |
| `Expected ">" but found "id"` | 9 | JSX markup inside a ```` ```ts ```` fence (should be `tsx`) |
| `Expected "}" but found "，"` | 4 | **full-width comma U+FF0C** in code — see below |
| `Expected "}" but found "smooth"` | 2 | **missing comma** — see below |
| `Unexpected end of file` | 2 | truncated block |

Affected files: `input-API`, `textarea-API`, `scroll-view-API`,
`scroll-coordinator-API`, `webview-API`, `viewpager-API`, `text`, `list`.

## Two independent, unambiguous defects found along the way

These are plain typos rather than convention, and can be fixed without deciding
anything:

1. **Full-width comma in `docs/zh/api/elements/built-in/image.mdx`** — lines 341,
   356, 371, 386 end `method: 'startAnimate'，` with U+FF0C instead of `,`.
   Pasting any of those four blocks is a syntax error.
2. **Missing comma in `list.mdx`** (`en:773`, `zh:757`, all branches):
   ```ts
   itemKey: `${itemKey}`
   smooth: true,
   ```

## The decision this needs

Review offered two representations. Both are defensible; the point is picking one
and applying it everywhere:

- **(a) Runnable values.** `params` carries concrete values (`value: 'hello'`,
  `smooth: true`). Reads as copy-paste code. Loses the ability to express
  optionality inline, which matters for `webview` `cookies.set` (9 fields, 7
  optional).
- **(b) Separated type + example.** Declare the shape as a type, then show a
  runnable invocation referencing it. Keeps every `@since` / platform annotation
  and expresses optionality, at the cost of two blocks per method.

(b) fits the `success: Callback<{...}>` cases better — those describe what the
callback *receives*, so there is no runtime value to show. Only three methods use
that form: `input.getValue`, `textarea.getValue`, `scroll-view.getScrollInfo`.

## Suggested landing plan

1. Agree the canonical representation.
2. Land it on `main` across all affected pages, en + zh together.
3. Cherry-pick to `release/4.0`, `3.9`, `3.8`, `3.7` in one series, so no
   published version disagrees with another — the inconsistency review raised.
4. Add the parser check to CI so new pages cannot reintroduce it.

## Why this is not in the current PRs

#1332/#1333 convert `viewpager-API` only, because an earlier review comment on
#1332 asked for exactly that, and it is the file those PRs already rewrite.
Extending them would pull in `text.mdx`, `list.mdx`, `image.mdx` and
`webview-API.mdx` — files untouched by the defect fixes those PRs exist for — and
would land a convention change on five release branches without coordination.

The measurement is reproducible:

```sh
node harness/check-example-syntax.mjs docs/en/api/elements/built-in/*.mdx
```

Shouqun:
Closes #1347.

```ts
params: {
  /** Input content */
```

The block is presented as something you can paste, and it does not parse.
Measured with esbuild's TypeScript parser over every fenced JS/TS block
containing a real `invoke({...})` call:

| | before | after |
| --- | --- | --- |
| `docs/en` | 20 of 43 failing | **0** |
| `docs/zh` | 28 of 52 failing | **0** |

## Representation

Two forms, chosen per block so nothing is lost:

- **All-required `params`** get concrete values inline, matching the
  `viewpager-API` page that landed in #1332. The per-field JSDoc stays attached.
  declared as an `interface` above the call and referenced from it:

```ts
interface ScrollToParams {
  /**
   * Target item index
   * @defaultValue 0
   */
  index?: number;
  ...
}

const params: ScrollToParams = { index: 0, offset: 0, smooth: true };

lynx.createSelectorQuery().select('#id').invoke({ method: 'scrollTo', params, ... }).exec();
```

Writing `name: 'session'` in place of `name?: string` would silently drop the
fact that a field is optional. That matters most for `webview` `cookies.set`,
which has nine fields, seven of them optional.

## Defects fixed along the way

Five unrelated problems surfaced in the same blocks:

- **Full-width commas (U+FF0C)** used as code punctuation on the Chinese pages —
  `image.mdx`, `list.mdx`, `scroll-view.mdx`, `view.mdx` and others.
- **A missing comma** in `list.mdx`: `` itemKey: `${itemKey}` `` followed
  directly by `smooth: true,`.
- **A garbled JSX tag** in the Chinese `scroll-view.mdx`:
  ``<`<scroll-view>`id="scroll"/>``.
- **An unterminated arrow function** in the cross-text-selection example in
  `text.mdx`.
- **Wrong fence languages** — JSX markup and TypeScript annotations inside
  fences declared `ts` or `jsx`, which need `tsx`.

The Chinese `selector-query-select` page also still described the failure case
with class-component pseudocode (`componentDidMount`, `should_render`) while the
English page had moved to hooks, and its example did not parse. It now matches
the English one, with the surrounding Chinese prose updated to refer to
`useEffect()` and `show`.

## Verification

The argument values and declared result shapes are not guesses. A ReactLynx page
calls the documented `<scroll-view>` methods with exactly the values now shown
and checks the results against the declared types, built and run on **Android
4.0.0** and **iOS 4.0.0**:

| Call | Android | iOS |
| --- | --- | --- |
| `scrollTo({ index: 0, offset: 0, smooth: true })` | accepted | accepted |
| `scrollBy({ offset: 100 })` | accepted, view moves | accepted, view moves |
| `getScrollInfo` result shape | `x=0 y=99 range=271` | `x=0 y=100 range=340` |

`getScrollInfo` returns exactly `scrollX` / `scrollY` / `scrollRange` — the
fields now declared as `GetScrollInfoResult`.

**macOS** was covered too: the app built from this repo's macOS integration
guide loads the same page over HTTP and renders it without error, and all five
Objective-C++ samples in that guide still compile against the released macOS SDK
headers.

Local CI — `check-case-collisions`, `check-asset-urls`, `check-lynx-ui-routes`,
`pnpm install --frozen-lockfile`, `pnpm run format:check`, `pnpm run build` —
passes on this branch.

`autoScroll` example shows `success` and `fail` callbacks, and **neither is
invoked** on Android or iOS, for `start: true` or `start: false`. If that is
intended, the example may be better off without them.

- Normalize Element API `invoke()` examples across English and Chinese documentation.
- Use concrete values for runnable parameters and named interfaces for optional fields and callback results.
- Resolve syntax defects, malformed JSX, truncated blocks, and incorrect code-fence languages.
- Align the Chinese `selector-query-select` example with the English hooks-based implementation.
- Verify 43/43 English and 52/52 Chinese `invoke()` blocks parse successfully.

---

## Review follow-up

**The expired cookie was a real defect and is fixed.** `cookies.set` carried
`expirationDate: 1767225600` — 2026-01-01, eight months in the past — so a
copied example would have expired the cookie immediately rather than
demonstrating a persistent one. It is now computed rather than hardcoded, so it
cannot go stale again:

```ts
expirationDate: Math.floor(Date.now() / 1000) + 60 * 60 * 24 * 30
```

EN and ZH are identical.

**The Chinese `scrollBy` placeholder is also fixed, and the note about it was
half right.** There is no `offset: number` left in `scroll-view-API.mdx` — this
PR converted that one. The surviving instance is in `scroll-view.mdx`, a
different page, and it is real:

```ts
offset: number, // 滚动的距离, 单位 px
```

Auditing that pattern found **nine** across `list.mdx` (both locales) and the
Chinese `scroll-view.mdx`: `rate: string,`, `start: bool,`, `autoStop: bool,`,
`offset: number,`. All are now values.

Worth stating why neither review caught these: **they parse.** `number` is a
valid JavaScript identifier, so `offset: number,` is syntactically fine and no
parser will ever object — it is only meaningless as documentation. The
syntax-based measurement in this PR is blind to it by construction, which is
also why they were not in the original count.

**One more value fixed while auditing:** `scrollBy` passed `offset: 0`, which
scrolls nothing. It now passes `100`, the value actually exercised on device.

Re-verified at this head: 0 failing examples in `docs/en` and `docs/zh`, 0 bare
type names remaining, Prettier clean, and the local CI job — including
`pnpm run build` — passing.

### DIFF
--- a/docs/en/api/elements/built-in/input-API.mdx
+++ b/docs/en/api/elements/built-in/input-API.mdx
@@ -263,49 +263,50 @@ Require focus
 
  <AndroidOnly /> <IOSOnly /> <HarmonyOnly /> <VersionBadge>3.4</VersionBadge>
 ```ts
+interface GetValueResult {
+  /**
+   * Is composing or not, iOS only
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  isComposing: boolean;
+  /**
+   * End position of the selection
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  selectionEnd: number;
+  /**
+   * Begin position of the selection
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  selectionStart: number;
+  /**
+   * Input content
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  value: string;
+}
 
 lynx.createSelectorQuery()
      .select('#id')
      .invoke({
       method: 'getValue',
-      success: Callback<{
-        /**
-         * Is composing or not, iOS only
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        isComposing: boolean;
-        /**
-         * End position of the selection
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        selectionEnd: number;
-        /**
-         * Begin position of the selection
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        selectionStart: number;
-        /**
-         * Input content
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        value: string;
-      }>;
+      success: function (res: GetValueResult) {},
       fail: function (res) {},
     })
     .exec();
@@ -333,7 +334,7 @@ lynx.createSelectorQuery()
          * @Web
          * @since 3.4
          */
-        selectionEnd: number;
+        selectionEnd: 5,
         /**
          * Start position of the selection
          * @Android
@@ -342,8 +343,8 @@ lynx.createSelectorQuery()
          * @Web
          * @since 3.4
          */
-        selectionStart: number;
-      };
+        selectionStart: 0,
+      },
       success: function (res) {},
       fail: function (res) {},
     })
@@ -372,8 +373,8 @@ lynx.createSelectorQuery()
          * @Web
          * @since 3.4
          */
-        value: string;
-      };
+        value: 'hello',
+      },
       success: function (res) {},
       fail: function (res) {},
     })
--- a/docs/en/api/elements/built-in/list.mdx
+++ b/docs/en/api/elements/built-in/list.mdx
@@ -770,7 +770,7 @@ this.createSelectorQuery()
       position: 10,
       offset: 100,
       alignTo: 'top',
-      itemKey: `${itemKey}`
+      itemKey: `${itemKey}`,
       smooth: true,
     },
     success: function (res) {},
@@ -797,9 +797,9 @@ this.createSelectorQuery()
   .invoke({
     method: 'autoScroll',
     params: {
-      rate: string, //  The distance scrolled per second, supports positive and negative. Distance supports units "px/rpx/ppx" default->null (iOS value must be greater than 1/screen.scale px)
-      start: bool, //  Start/pause auto-scrolling default->false
-      autoStop: bool, // Whether to automatically stop when reaching the bottom default->true
+      rate: '60px', //  The distance scrolled per second, supports positive and negative. Distance supports units "px/rpx/ppx" default->null (iOS value must be greater than 1/screen.scale px)
+      start: true, //  Start/pause auto-scrolling default->false
+      autoStop: true, // Whether to automatically stop when reaching the bottom default->true
     },
     success: function (res) {},
     fail: function (res) {},
@@ -858,7 +858,7 @@ lynx
   .invoke({
     method: 'scrollBy',
     params: {
-      offset: number,
+      offset: 100,
     },
     success(res) {
       console.log('succ ');
--- a/docs/en/api/elements/built-in/scroll-coordinator-API.mdx
+++ b/docs/en/api/elements/built-in/scroll-coordinator-API.mdx
@@ -153,7 +153,7 @@ lynx.createSelectorQuery()
          * @Web
          * @PC
          */
-        offset: string;
+        offset: '0px',
         /**
          * Whether folding requires animation
          * @Android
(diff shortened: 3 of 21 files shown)


Huxpro (changes requested):
One concrete defect introduced by this PR needs fixing before merge: the new `cookies.set` examples use `expirationDate: 1767225600`, which is January 1, 2026 and therefore already expired on August 25, 2026. A copied example will immediately expire the cookie instead of demonstrating a successful persistent cookie. Please either omit this optional field or compute a future timestamp, and keep EN/ZH aligned. The existing inline thread already points to the exact lines, so I’m not duplicating it.

I’m not blocking on the pre-existing `offset: number` runtime placeholder in the Chinese `scrollBy` example; that can be handled as a follow-up because this PR did not introduce it. CI is green on the current head.

This is a Codex-assisted automated review. Please contact me in person for a second opinion.

Huxpro (approved):
Re-reviewed the latest head after the follow-up fixes. The expired cookie example is now computed relative to `Date.now()`, the additional bare type-name placeholders and missing comma are corrected, EN/ZH stay aligned, and the exact head has successful CI and deploy-preview checks. I do not see a concrete defect introduced by the current diff.

PR #1351, opened 2026-08-25

Shouqun:
Backport of #1350 to `release/3.7`, so the same page does not disagree with
itself depending on the selected site version. Part of #1347.

```ts
params: {
  /** Input content */
```

The block is presented as something you can paste, and it does not parse. Every
such block on this branch now does — measured with esbuild's TypeScript parser
over every fenced JS/TS block containing a real `invoke({...})` call.

## Representation

- **All-required `params`** get concrete values inline; the per-field JSDoc
  stays attached.
  declared as an `interface` above the call and referenced from it. Writing
  `name: 'session'` in place of `name?: string` would silently drop the fact
  that a field is optional.

## Defects fixed in the same blocks

- **Full-width commas (U+FF0C)** used as code punctuation on the Chinese pages.
- **A garbled JSX tag** — `<`<scroll-view>`id="scroll"/>`.
- **An unterminated arrow function** in the cross-text-selection example.
- **Bare type names used as values** — `rate: string,`, `offset: number,`.
  These *parse* as JavaScript, since `number` is a valid identifier, so no
  syntax check can see them; they are the same defect all the same.
- **Wrong fence languages** — JSX and TypeScript annotations in `ts`/`jsx`
  fences, which need `tsx`.

The Chinese `selector-query-select` page also still taught class components
(`componentDidMount`, `should_render`) where the English page had moved to
hooks, and its example did not parse. It now matches, with the Chinese prose
updated.

## Verification against this branch's own SDK

This branch pins **Lynx 3.7.0**. A ReactLynx page calls the documented
`<scroll-view>` methods with exactly the values now shown and checks results
against the declared types, run on **Android 3.7.0**:

| Call | Result |
| --- | --- |
| `scrollTo({ index: 0, offset: 0, smooth: true })` | accepted |
| `scrollBy({ offset: 100 })` | accepted, view moves |
| `getScrollInfo` | returns `scrollX` / `scrollY` / `scrollRange` — the fields now declared as `GetScrollInfoResult` |

All five Objective-C++ samples in this branch's **macOS** integration guide
still compile against the released macOS SDK headers, in both locales.

Syntax check on this head: `docs/en` 51 checked, 0 failed.

`autoScroll` example shows `success` and `fail` callbacks, and neither is
invoked, for `start: true` or `start: false`. Reproduced on every SDK line
tested.

### DIFF
--- a/docs/en/api/elements/built-in/input-API.mdx
+++ b/docs/en/api/elements/built-in/input-API.mdx
@@ -293,49 +293,50 @@ Require focus
 
 <AndroidOnly /> <IOSOnly /> <HarmonyOnly /> <VersionBadge>3.4</VersionBadge>
 ```ts
+interface GetValueResult {
+  /**
+   * Is composing or not, iOS only
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  isComposing: boolean;
+  /**
+   * End position of the selection
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  selectionEnd: number;
+  /**
+   * Begin position of the selection
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  selectionStart: number;
+  /**
+   * Input content
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  value: string;
+}
 
 lynx.createSelectorQuery()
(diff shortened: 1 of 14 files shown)


Huxpro (approved):
Reviewed the release/3.7 backport against the coordinated #1347 landing plan. The diff is limited to the affected EN/ZH example pages that exist on this branch, preserves the branch's API/version surface, and applies the same runnable-example corrections without introducing new APIs or compatibility claims. `release/3.7` does not have the normal CI workflow; the configured CLA/Netlify checks on this exact head are successful.

PR #1354, opened 2026-08-25

Shouqun:
Backport of #1350 to `release/4.0`, so the same page does not disagree with
itself depending on the selected site version. Part of #1347.

```ts
params: {
  /** Input content */
```

The block is presented as something you can paste, and it does not parse. Every
such block on this branch now does — measured with esbuild's TypeScript parser
over every fenced JS/TS block containing a real `invoke({...})` call.

## Representation

- **All-required `params`** get concrete values inline; the per-field JSDoc
  stays attached.
  declared as an `interface` above the call and referenced from it. Writing
  `name: 'session'` in place of `name?: string` would silently drop the fact
  that a field is optional.

## Defects fixed in the same blocks

- **Full-width commas (U+FF0C)** used as code punctuation on the Chinese pages.
- **A garbled JSX tag** — `<`<scroll-view>`id="scroll"/>`.
- **An unterminated arrow function** in the cross-text-selection example.
- **Bare type names used as values** — `rate: string,`, `offset: number,`.
  These *parse* as JavaScript, since `number` is a valid identifier, so no
  syntax check can see them; they are the same defect all the same.
- **Wrong fence languages** — JSX and TypeScript annotations in `ts`/`jsx`
  fences, which need `tsx`.

The Chinese `selector-query-select` page also still taught class components
(`componentDidMount`, `should_render`) where the English page had moved to
hooks, and its example did not parse. It now matches, with the Chinese prose
updated.

## Verification against this branch's own SDK

This branch pins **Lynx 4.0.0**. A ReactLynx page calls the documented
`<scroll-view>` methods with exactly the values now shown and checks results
against the declared types, run on **Android 4.0.0**, and on iOS 4.0.0:

| Call | Result |
| --- | --- |
| `scrollTo({ index: 0, offset: 0, smooth: true })` | accepted |
| `scrollBy({ offset: 100 })` | accepted, view moves |
| `getScrollInfo` | returns `scrollX` / `scrollY` / `scrollRange` — the fields now declared as `GetScrollInfoResult` |

All five Objective-C++ samples in this branch's **macOS** integration guide
still compile against the released macOS SDK headers, in both locales.

Syntax check on this head: `docs/en` 61 checked, 0 failed.

`autoScroll` example shows `success` and `fail` callbacks, and neither is
invoked, for `start: true` or `start: false`. Reproduced on every SDK line
tested.

---

## Review follow-up

**The regression was real and is fixed on this head.** `getVisibleCells`
documents the shape of what the call *returns*, so `index: 0` was wrong beside
`id: number` / `itemKey: string` / `left: number`. `index: number` is restored in
EN and ZH, and this PR no longer touches that line at all — it now matches
`main` exactly.

Cause: the value pass matched on field name across the whole file instead of
only inside call examples, so it reached a return-shape block. Auditing that
pattern across every affected branch found **exactly this one line per locale**
and nothing else, on 3.7, 3.8, 3.9 and 4.0.

The already-merged branches are covered by #1355 (`release/3.7`), #1356
(`release/3.8`) and #1357 (`release/3.9`), so the versioned copies stay aligned.

**Verified rather than assumed.** A ReactLynx page renders a `<list>` and calls
`getVisibleCells`, asserting the returned indices are numbers and that they
vary:

| Platform | Result |
| --- | --- |
| Android 4.0.0 | `index = [0, 1, 2]` |
| iOS 4.0.0 | `index = [0, 1, 2]` |

The macOS app built from this branch's integration guide renders the same page
over HTTP without error, and all five Objective-C++ samples still compile.

At this head: 0 failing `invoke()` examples in `docs/en` and `docs/zh`, Prettier
clean, one commit.

### DIFF
--- a/docs/en/api/elements/built-in/input-API.mdx
+++ b/docs/en/api/elements/built-in/input-API.mdx
@@ -263,49 +263,50 @@ Require focus
 
  <AndroidOnly /> <IOSOnly /> <HarmonyOnly /> <VersionBadge>3.4</VersionBadge>
 ```ts
+interface GetValueResult {
+  /**
+   * Is composing or not, iOS only
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  isComposing: boolean;
+  /**
+   * End position of the selection
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  selectionEnd: number;
+  /**
+   * Begin position of the selection
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  selectionStart: number;
+  /**
+   * Input content
+   * @Android
+   * @iOS
+   * @Harmony
+   * @Web
+   * @since 3.4
+   */
+  value: string;
+}
 
 lynx.createSelectorQuery()
      .select('#id')
      .invoke({
       method: 'getValue',
-      success: Callback<{
-        /**
-         * Is composing or not, iOS only
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        isComposing: boolean;
-        /**
-         * End position of the selection
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        selectionEnd: number;
-        /**
-         * Begin position of the selection
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        selectionStart: number;
-        /**
-         * Input content
-         * @Android
-         * @iOS
-         * @Harmony
-         * @Web
-         * @since 3.4
-         */
-        value: string;
-      }>;
+      success: function (res: GetValueResult) {},
       fail: function (res) {},
     })
     .exec();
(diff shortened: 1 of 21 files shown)


Huxpro (approved):
Reviewed the release/4.0 backport against the final #1350 head. The previously fixed cookie-expiration issue is carried over as a runtime-computed future timestamp, the EN/ZH examples remain aligned, and no API/compatibility/version metadata is changed. CI and the branch's deploy/CLA checks are green on this exact head.

Huxpro (changes requested):
A concrete branch-specific regression remains on this head. In the `getVisibleCells` return-shape example, the PR changes only `index: number` to `index: 0` while every neighboring field remains a type placeholder (`id: number`, `itemKey: string`, `left: number`, etc.). That block documents the returned cell shape, so `index: 0` incorrectly implies the index is the literal value 0 rather than a number. Please restore `index: number` in both EN and ZH (or convert the entire block to an explicitly concrete example, which would be a larger change).

This same backport-only change was also carried into #1351–#1353; main/#1350 does not contain it. I will track those already-landed release branches separately so the versioned copies stay aligned.

This is a Codex-assisted automated review. Please contact me in person for a second opinion.

Huxpro (approved):
Re-reviewed the latest head. The prior getVisibleCells regression is fixed in both EN and ZH, the only follow-up delta restores index: number, and the remaining backport changes match the merged #1350 changes modulo release/4.0 baseline differences. Validate, deploy, and formatting checks are green. No remaining actionable findings.

PR #1371, opened 2026-08-26
docs(api): generate Element API reference from Lynx source

Huxpro:
## Summary

- generate the Element API reference with TypeDoc directly from the declarations in `lynx-family/lynx`
- export only the 42 APIs in the current public Element API reference and fail generation when that surface drifts
- exclude internal renderer APIs and preserve the existing public URLs, compatibility metadata, and bilingual navigation
- extend the source-driven API update workflow to check out both `lynx-stack` and `lynx`

This replaces the stale package-based Element API output with documentation derived from the runtime-aligned declaration source.

## Validation

- targeted Element API TypeDoc generation is reproducible with a clean worktree
- 42 English and 42 Chinese public function pages are generated
- full Rspress build succeeds and writes 3,197 sitemap pages
- only `/api/engine/element-api.html` and `/zh/api/engine/element-api.html` are emitted as overview routes; no duplicate `index.html` routes are produced
- canonical URLs for both locales are correct

## Dependency

The corresponding `type-element-api` declaration and JSDoc changes must land in `lynx-family/lynx` `develop` before the scheduled regeneration workflow is used.


Huxpro (commented):
Blocking findings — REQUEST_CHANGES is unavailable because this account owns the PR.

Codex-assisted automated review on the current head found two concrete source-of-truth/workflow blockers:

1. The Lynx checkout path is wrong in both regeneration paths. The workflow checks out `lynx-family/lynx` at `.lynx`, and the declaration in that repository is at `js_libraries/type-element-api/types/element-api.d.ts`. Both `scripts/update-api-docs.sh` and `scripts/typedoc/generate-element-api-entry.ts` currently look for `lynx/js_libraries/type-element-api/types/element-api.d.ts`, adding a nonexistent top-level `lynx/` directory. A scheduled/manual regeneration using the workflow as written will therefore exit with “Element API declaration not found”. Please point both places at `js_libraries/type-element-api/types/element-api.d.ts` and validate against a normal checkout root.

2. The PR’s explicit upstream dependency has not landed yet. `lynx-family/lynx#5658` is still open, while this workflow immediately starts regenerating from `develop`. Current `develop` does not yet have the JSDoc/`@internal` source shape this generator expects; with the 42-function exact-surface guard, regeneration will fail on the extra global functions once the path above is corrected. Please land the corresponding Lynx source change first (or otherwise pin/guard the workflow so it cannot run against the incompatible source), then rebase/regenerate from the landed canonical declaration.

The regular website CI being green does not exercise this scheduled cross-repo regeneration path, so it does not cover either failure mode. Please contact me in person for a human second opinion.

Huxpro (commented):
Additional blocking finding on the same Codex-assisted review: the generated Chinese Element API is not actually bilingual. The PR runs the same English JSDoc entry for `zh`, so existing Chinese descriptions are replaced by English text, and generated cross-links/breadcrumbs use `/api/...` instead of `/zh/api/...`. For example, the Chinese `__CreateElement` page changes “创建 Element 对象” and the Chinese parameter descriptions to English, and links `ElementInfo`/`ElementRef` into the English API tree.

That is a concrete locale regression introduced by this PR, not a pre-existing translation gap. Please preserve Chinese content/cross-links (for example via a locale-aware overlay/source mechanism) before replacing the existing hand-written ZH reference. Please contact me in person for a human second opinion.

PR #1377, opened 2026-08-26
fix(docs): Clay does not support scroll-view scrollBy/getScrollInfo o…

MoonfaceX:
…r accessibilityAnnounce/cancelResourcePrefetch

PR #1363 only removed clay_macos support, but Clay Windows also does not support these APIs. Additionally, scroll-view's scrollBy and getScrollInfo are not supported on any Clay platform (not just macOS).

Changes:
- scroll-view-API: remove all Clay badges from scrollBy and getScrollInfo
- compat data: set clay_windows to false for accessibilityAnnounce and cancelResourcePrefetch

- Remove Clay support badges from `scrollBy` and `getScrollInfo` documentation.
- Mark `accessibilityAnnounce` and `cancelResourcePrefetch` as unsupported on Clay Windows.

### DIFF
--- a/docs/en/api/elements/built-in/scroll-view-API.mdx
+++ b/docs/en/api/elements/built-in/scroll-view-API.mdx
@@ -196,7 +196,7 @@ Automatic scrolling
 
 ### `getScrollInfo`
 
- <AndroidOnly /> <IOSOnly /> <ClayWindowsOnly /> <HarmonyOnly />
+ <AndroidOnly /> <IOSOnly /> <HarmonyOnly />
 ```ts
 interface GetScrollInfoResult {
   /**
@@ -241,7 +241,7 @@ Get scroll info
 
 ### `scrollBy`
 
- <AndroidOnly /> <IOSOnly /> <ClayWindowsOnly /> <HarmonyOnly />
+ <AndroidOnly /> <IOSOnly /> <HarmonyOnly />
 ```ts
 interface ScrollByParams {
   /**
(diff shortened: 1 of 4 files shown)


Shouqun (approved):

Huxpro:
## Maintainer decision needed

I found a source/runtime-contract conflict after this merged. The Clay implementation in `clay/ui/component/scroll_wrapper.cc` still registers `getScrollInfo` unconditionally in `LYNX_UI_METHOD_BEGIN(ScrollWrapper)` and returns `scrollTop`, `scrollLeft`, `scrollHeight`, `scrollWidth`, and `isDragging`; `builtin_views.cc` also registers `scroll-view`/`x-scroll-view` with `ScrollWrapper` without a macOS/Windows guard. That is direct evidence that the common Clay code path exposes the method. By contrast, `scrollBy` is not registered there, so removing Clay for `scrollBy` looks consistent with source.

Risk: #1377 removed the remaining `ClayWindowsOnly` badge for `getScrollInfo`, so the published API now says no Clay platform supports a method that the shared Clay implementation appears to expose. Earlier runtime testing may have found a platform/build-specific failure, so I do not think the source conflict alone is enough to assert the runtime claim is definitely wrong.

Options: (1) re-run `getScrollInfo` against the exact current Clay Windows/macOS artifacts and keep the badges removed if it genuinely fails, documenting why the registered common implementation is unreachable; or (2) restore the appropriate Clay badge(s) if the current builds expose the registered method. My recommendation is to re-verify Windows first, because this PR specifically removed a previously documented Windows capability while the common source still implements it.

This is a Codex-assisted review. Please contact Xuan in person for a human second opinion.

PR #1381, opened 2026-08-26
[Doc] Generate public Element compatibility data

liuzhenyue-oss:
[Doc] Generate public Element compatibility data

- Regenerate public Element compatibility JSON from TypeScript declarations, including surface-specific overrides and nested Element projections.
- Store Web support as boolean values verified against the lynx-stack Web implementation instead of assigning Lynx release versions.
- Apply Element-root Web availability to child APIs and keep generated JSON key order stable.
- Preserve the existing Element support corrections, independent surface files, and android-overlay-scope metadata.
- Keep this change limited to generated compatibility data and its ignore rule.

TEST: With pull request #1380 applied, full compatibility generation has zero unresolved paths; repeated generation has zero add, update, or delete operations; parsed all 39 JSON files; verified zero string or array Web support statements

Change-Id: Iba4eb2d4e036d5e2cc2a159a83945270477b8455

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,9 @@ lynx-examples/
 packages/lynx-compat-data/css/properties/*.json
 packages/lynx-compat-data/api-stats.json
 
+# Element compat source indexes are regenerated on demand by tooling.
+packages/lynx-compat-data/.generated/
+
 !*logger.md
 .reference/
 .trae/
--- a/packages/lynx-compat-data/elements/blur-view.json
+++ b/packages/lynx-compat-data/elements/blur-view.json
@@ -91,6 +91,9 @@
               },
               "web_lynx": {
                 "version_added": false
+              },
+              "clay_harmony": {
+                "version_added": false
               }
             }
           }
@@ -125,6 +128,9 @@
                 "version_added": false
               },
               "web_lynx": {
+                "version_added": false
+              },
+              "clay_harmony": {
                 "version_added": false
               }
             }
@@ -161,6 +167,9 @@
               },
               "web_lynx": {
                 "version_added": false
+              },
+              "clay_harmony": {
+                "version_added": false
               }
             }
           }
@@ -195,6 +204,9 @@
                 "version_added": false
               },
               "web_lynx": {
+                "version_added": false
+              },
+              "clay_harmony": {
                 "version_added": false
               }
             }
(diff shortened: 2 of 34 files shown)


Huxpro (dismissed):
I’m requesting changes on this exact head for a concrete compatibility-data inconsistency that makes the generated dataset unsafe to land as-is. I verified the existing inline finding in `elements/frame.json`: `frame.__compat` still marks every Clay surface unsupported, but the regenerated `frame.bindload` entry claims Clay Android/iOS/macOS/Windows/Harmony support since 3.3. A child API cannot be supported on a platform where the element itself is unavailable, so at least this generated record is internally contradictory. Please fix the generator/source metadata (rather than hand-editing generated output) and regenerate the affected files. I’m not duplicating the existing inline comments for the other generated inconsistencies already identified on this head.

This PR also currently depends on #1380: the regenerated JSON contains `clay_harmony`, which the target branch schema/types do not accept until #1380 lands. Please keep the merge order atomic and rerun the normal CI + Compat Data Stats workflows after #1380 is present in the base.

This is a Codex-assisted review. Please contact me in person for a second opinion if you think the generated source semantics justify the current result.

Huxpro (dismissed):
Latest-head re-review: the previous `frame.bindload` inconsistency is fixed, so I dismissed that outdated review.

One concrete merge-order blocker remains. This head still generates `clay_harmony` compatibility entries and the PR explicitly says validation was performed with #1380 applied, but #1380 is still unmerged and `main` does not yet have that schema/platform support. Merging #1381 first would land generated data that depends on a base contract the target branch does not yet contain. Please land #1380 first, then rebase/regenerate #1381 against the updated base and rerun the normal CI + Compat Data Stats checks on the final head. On this exact SHA, both workflows currently end as `action_required` rather than executing.

I’m not duplicating the already-open generated-data review threads on this head.

This is a Codex-assisted review. Please contact me in person for a second opinion if you think the dependency ordering can safely be relaxed.

Huxpro (approved):
Latest-head re-review: #1380 has landed and this PR is now based on that schema/platform change. The incremental regeneration also resolves the previously raised grouped Web-support and duplicated bind-event findings, and the remaining review threads are resolved. I found no new concrete compatibility-data defect on this head. Approval is for the content; merge should still wait for the configured CI / Compat Data Stats workflows to actually run and for the CLA requirement to clear.

PR #1385, opened 2026-08-26
Revert "docs(react): clarify Snapshot runtime pseudocode (#1349)"

Huxpro:
Reverts #1349 because the follow-up changed substantially more of the section than intended. This restores both English and Chinese documents exactly to their state before #1349 while retaining #1277.

- Restore the English and Chinese compiler-informed rendering documentation to its pre-PR `#1349` structure.
- Retain PR `#1277` while restoring the formal pseudocode types and runtime terminology.

### DIFF
--- a/docs/en/react/compiler-informed-rendering.mdx
+++ b/docs/en/react/compiler-informed-rendering.mdx
@@ -256,93 +256,33 @@ The `$0` and `__DynamicPartSlotV2` representation shown here reflects `@lynx-js/
 
 One characteristic of ReactLynx is that this compiled representation participates not only in update-time reconciliation, but also in Lynx's dual-thread [Instant First-Frame Rendering (IFR)](/guide/interaction/ifr) pipeline.
 
-#### Concepts Used in the Following Pseudocode
-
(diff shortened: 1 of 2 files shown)


Huxpro (commented):
Codex-assisted review — please contact me in person for a second opinion.

Blocking finding (GitHub does not allow me to formally request changes on my own PR): this revert reintroduces a concrete inconsistency between the two adjacent pseudocode snippets. The compiled updater shown immediately above reads `ctx.__elements` and `ctx.__values`, but the restored first-render/update pseudocode writes `instance.elements` and `instance.values` before invoking that same `definition.update[...]` function. Taken together as written, the updater will not see the values/elements the pseudocode just populated. The runtime source confirms `SnapshotInstance` actually stores these fields as `__elements` / `__values`, and `Snapshot.update` receives a `SnapshotInstance`.

#1349 avoided this by using explicitly separate explanatory types/field names and stating that they were conceptual rather than the implementation fields. If the intent is to simplify/reduce #1349, please keep that abstraction boundary (or make the restored pseudocode internally consistent) rather than reverting to snippets that do not compose.

Huxpro:
@coderabbitai resolve

PR #1406, opened 2026-09-04
docs(react): clarify compiler and runtime collaboration

Huxpro:
## Summary

- reuse the conceptual Greeting template when explaining the initial render
- describe how main-thread and background-thread template instances correspond
- replace Snapshot-specific indices and Patch encoding with dynamic-data pseudocode
- keep the boundary clear between targeted property updates and reconciled dynamic children

## Checks

- git diff --check
- Prettier check for both modified MDX files
- CSpell for both modified MDX files

- Rename “First-screen Rendering” to “First-frame Rendering” and define Instant First-Frame Rendering (IFR).
- Clarify compiler and runtime collaboration through Greeting template instances, hydration, and dynamic updates.
- Replace Snapshot-specific patch details with dynamic-data pseudocode and update performance references and living specifications.


Huxpro (commented):
Maintainer decision needed

Codex-assisted review. The two incremental commits rename the living-spec concept from **First-screen Rendering / FSR** to **First-frame Rendering** and change the published spec anchor from `#first-screen-rendering-or-fsr` to `#first-frame-rendering`. The branch updates the known in-repo links it touches, and repository CI plus the Netlify preview pass, so I do not have evidence for a blocking correctness defect.

There are two merge-risk details that need an explicit maintainer choice:

- The old public anchor is no longer emitted. That means existing external/bookmarked deep links to `/guide/spec.html#first-screen-rendering-or-fsr` will stop landing on the section even though in-repo references are updated.
- Regenerating `docs/public/living-spec/index.html` also brings unrelated Bikeshed boilerplate drift (for example `// TODO: shared util`) that is not present in the spec source or the current checked-in HTML. #1401 exists specifically because the current generator/spec-data inputs are not reproducible yet. Separately, user-facing docs such as `docs/en/react/lifecycle.mdx` and the generated `createPortal` API reference still use “first-screen rendering,” so merging this creates a terminology transition rather than an atomic rename.

**Options:** (1) preserve the old anchor as a compatibility alias and leave broader wording alignment as follow-up; (2) align the same-concept docs atomically; or (3) intentionally accept the route/terminology break and current generator drift. **Recommendation:** preserve the old anchor for compatibility, and preferably regenerate after #1401 (or otherwise keep the generated HTML diff limited to the intended semantic change). I am not blocking the PR on this evidence alone.

Please contact me in person for a second opinion before deciding.

Huxpro (commented):
Maintainer decision needed

Codex-assisted review of exact head `a688ec63cbccf7aaa59979a1c04218ddfdf64d36`, superseding my earlier gray-area note where it is now stale.

The new incremental commit only changes the EN/ZH hydration/IFR explanation. I validated the mechanism against the current ReactLynx runtime: the main-thread root is a `SnapshotInstance`; `renderPage()` creates the page/elements and runs the main-thread render before first-screen synchronization; the background runtime then receives the serialized main-thread tree, hydrates its `BackgroundSnapshotInstance` tree against it, and commits the hydration diff. The branch's Living Spec likewise defines First-frame Rendering as main-thread rendering that blocks the UI thread until the first frame completes. So the new “main thread first, background thread hydrates/attaches later” framing and IFR terminology are consistent with the implementation and spec. I found no concrete defect introduced by this increment.

The generator-drift concern from my previous review is also obsolete now that #1401 has merged and makes Living Spec generation reproducible. The remaining gray area is public-anchor compatibility: this PR changes the published section id from `#first-screen-rendering-or-fsr` to `#first-frame-rendering` and does not preserve the old id. In-repo references are updated and the preview is green, so I do not have evidence to block the PR, but existing external/bookmarked deep links may stop landing on the section.

**Options:** intentionally accept the anchor transition, or preserve `#first-screen-rendering-or-fsr` as a compatibility alias while using the new First-frame Rendering terminology. **Recommendation:** preserve the old anchor as an alias if Bikeshed can express it without reintroducing generated drift; otherwise make the compatibility break an explicit maintainer decision.

Please contact me in person for a second opinion before merging.

PR #1407, opened 2026-09-05
docs(rspeedy): build an external bundle of plain TypeScript

upupming:
Documents building an external bundle of plain TypeScript, without `pluginReactLynx`.

Depends on lynx-family/lynx-stack#3814, which makes `pluginLynx` supply the runtime wrapper and the encoder such a bundle needs, and exports the `LAYERS` an entry picks a thread with. The section also covers declaring the external for one thread only — the string shorthand declares it for both, which fails to load a bundle built for one.

- Document plain TypeScript external bundles with `pluginLynx` and `LAYERS`.
- Explain single-thread and both-thread bundle outputs.
- Show matching external declarations in English and Chinese examples.

### DIFF
--- a/docs/en/rspeedy/external-bundle.mdx
+++ b/docs/en/rspeedy/external-bundle.mdx
@@ -72,6 +72,48 @@ After a successful build, you will get an output directory like this:
 │   └── comp-lib.template.js
 ```
 
+### Without A DSL
+
+A bundle of plain TypeScript needs no DSL plugin. `pluginLynx` is the Lynx build engine, and `LAYERS` picks the thread the bundle runs on:
+
+```ts title="rslib-utils.config.ts"
+import {
+  defineExternalBundleRslibConfig,
+  LAYERS,
+} from '@lynx-js/lynx-bundle-rslib-config';
+import { pluginLynx } from '@lynx-js/rsbuild-plugin';
+
+export default defineExternalBundleRslibConfig({
+  id: 'utils',
+  source: {
+    entry: {
+      './utils.js': {
+        import: './src/utils.ts',
+        layer: LAYERS.BACKGROUND,
+      },
+    },
+  },
+  plugins: [pluginLynx()],
+});
+```
+
+An entry without a `layer` is emitted for both threads. Pick one when the host app only calls into the bundle from that thread: the bundle then carries a single section, and the host declares the external for that thread only.
+
+```ts title="lynx.config.ts"
+pluginExternalBundle({
+  externals: {
+    './utils.js': {
+      bundlePath: 'utils.lynx.bundle',
+      libraryName: './utils.js',
+      background: { sectionPath: './utils.js' },
+      async: true,
+    },
+  },
+});
+```
+
+The string shorthand (`'./utils.js': 'utils.lynx.bundle'`) declares the external for both threads, so a bundle built for one of them fails to load on the other.
+
 ### Bundle CSS
 
 <VersionBadge v={3.7} />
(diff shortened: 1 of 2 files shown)


Huxpro (dismissed):
Codex-assisted review.

Two concrete blockers in the new “Without A DSL” section:

1. This documents a toolchain contract that is not available on the target branch / released packages yet. `lynx-family/lynx-stack#3814` is still open, and that PR is specifically what exports `LAYERS` from `@lynx-js/lynx-bundle-rslib-config` and makes `pluginLynx()` provide the runtime wrapper/encoder for Rslib. Current `lynx-stack` main (`@lynx-js/lynx-bundle-rslib-config` 0.8.0) does not export `LAYERS`, and the current generated website API reference has no `LAYERS` entry. Merging this now would make the canonical guide recommend a config that cannot compile/run with the shipped contract. Please land and publish #3814 first and regenerate the API reference (or otherwise gate this section to a prerelease that actually contains the feature) before merging these instructions.

2. The new build snippet directly imports `pluginLynx` from `@lynx-js/rsbuild-plugin`, but the page’s Install command in both locales still omits that package. The upstream implementation PR itself adds `@lynx-js/rsbuild-plugin` to the example devDependencies. Please add it to the documented install set so the example does not rely on an undeclared/transitive dependency.

I’m not treating the existing CodeRabbit fragment-context comment as an additional blocker here. Please contact Xuan in person for a second opinion if you think either item should be handled differently.

Huxpro (dismissed):
Codex-assisted review.

The incremental commit fixes the missing `@lynx-js/rsbuild-plugin` install, and `lynx-family/lynx-stack#3814` has now merged, so I dismissed my previous review as stale. One concrete blocker remains: the page now tells readers to install the packages without a version and immediately use the new `LAYERS` / plain-`pluginLynx()` contract, but that contract has not reached the stable package/API surface yet.

The currently tagged stable releases are still `@lynx-js/lynx-bundle-rslib-config@0.8.0` and `@lynx-js/rsbuild-plugin@0.1.0`; both tags point to the Sep 4 release commit, which predates the Sep 5 merge of #3814. #3814 itself adds patch changesets for both packages and introduces the public `LAYERS` export. The current website API-reference source also has no `LAYERS` entry. As written, a reader following the unversioned install command can therefore receive stable packages that do not contain the documented API/behavior.

Please publish the patch releases containing #3814 and regenerate/land the API reference (and state the minimum supported versions if that is not otherwise unambiguous), or gate/pin this section to a prerelease that actually contains the feature.

Please contact Xuan in person for a second opinion if you think this publication/API alignment should be handled differently.

HuJean (approved):

Huxpro (changes requested):
Codex-assisted review.

The package-publication part of my previous review is now stale: lynx-stack release PR #3810 has merged with `@lynx-js/lynx-bundle-rslib-config@0.8.1` and `@lynx-js/rsbuild-plugin@0.1.1`, and its release notes explicitly include the plain-`pluginLynx()` / exported-`LAYERS` work from #3814.

One concrete blocker remains on this exact head: the guide now introduces `LAYERS` as the public API readers should import, but the current website API reference for `@lynx-js/lynx-bundle-rslib-config` still does not list `LAYERS` (or its `ExposedLayers` type), while the current upstream API Extractor report marks both as `@public`. That leaves the newly documented API absent from the canonical reference surface.

Please regenerate and land the `@lynx-js/lynx-bundle-rslib-config` API reference from the released/current upstream API before merging this guide, so the guide and API reference land in alignment.

Please contact Xuan in person for a second opinion if you think this API-reference alignment should be handled differently.

docs/en/rspeedy/external-bundle.mdx
```diff
@@ -72,6 +72,50 @@ After a successful build, you will get an output directory like this:
 │   └── comp-lib.template.js
 ```
 
+### Without A DSL
```
Huxpro, line 75: We don't use word DSL in official docs.

Huxpro (changes requested):

upupming:
@Huxpro Comments resolved, it there any problems?

PR #1409, opened 2026-09-06
docs: refresh the 4.0 SDK version and stop quick-start hardcoding it

Shouqun:
`docs/public/version.json` is not a changelog. `LYNX_VERSION` is interpolated straight into the integration guides' Gradle and Podfile snippets, so whatever it says is what every new integrator copies into their build. It has been sitting on `4.0.0` since 4.0.1 shipped on 2026-07-31 and 4.0.2 on 2026-09-04.

### What changed

| | before | after |
| --- | --- | --- |
| `LYNX_VERSION` | `4.0.0` | `4.0.2` |
| `versions[4.0].release_number` | `4.0.0` | `4.0.2` |
| quick-start download links (en + zh) | hardcoded `4.0.0` | `versionJson.LYNX_VERSION` |

`PRIMJS_VERSION` is unchanged and still correct: `4.0.0` is the newest stable in its line — 4.0.1 exists only as alphas, and 4.1.x is a different line, which is a maintainer's call rather than a refresh.

### Why quick-start is interpolated rather than retyped

The macOS and Windows fragments in this same directory already build the release URL from the manifest:

```jsx
<Link href={'https://github.com/lynx-family/lynx/releases/tag/' + versionJson.LYNX_VERSION}>
```

quick-start hardcoded the version in five places instead, which is how it drifted two patches behind while its siblings stayed current. Pointing it at the same source of truth means the next release bump is one edit, not six, and this page cannot silently fall behind again. All four assets it links exist under identical names in 4.0.2 (`LynxExplorer-arm64.app.tar.gz`, `LynxExplorer-x86_64.app.tar.gz`, `LynxExplorer-noasan-release.apk`, `lynx_explorer-default-unsigned.hap`).

### Verification

The version bump is not just "a newer number is published" — the guides' own instructions were followed at the new version:

- Built the Android host app from `guide/start/fragments/android/integrating-lynx-with-existing-app-android` with `lynxVersion=4.0.2`. Every coordinate resolves — `lynx`, `lynx-jssdk`, `lynx-trace`, `lynx-devtool`, the five `lynx-service-*`, `service-api`, and all ten `xelement-*` artifacts at 4.0.2 — paired with `primjs:4.0.0` exactly as the manifest specifies, producing a working debug APK. That is the pairing this change asks readers to use.
- 4.0.2 is a normal GitHub release: `prerelease=false`, `draft=false`, published 2026-09-04.
- `pnpm build` renders 3190 pages; the built HTML carries `releases/download/4.0.2/…` and `releases/tag/4.0.2` in **both** locales.
- `format:check`, `cspell:check` (2443 files), `check-docs` (1725 files) and the five `scripts/ci` checks pass.
- A full four-platform conformance run (web / Android / iOS / Clay-macOS, three rounds each) reports no new findings against its baselines.

### Two related things left alone

**Older lines in the picker.** `versions[3.9].release_number` says `3.9.0` while 3.9.1 shipped 2026-08-25, and `versions[3.3]` says `3.3.2` while 3.3.3 shipped 2025-11-13. Same pattern, but those entries label documentation built from `release/3.9` and `release/3.3`, so whether the label should track the newest patch or the patch that branch documents is a maintainer's call, not mine. Happy to include them if you want them refreshed.

**The Harmony fragment** pins `@lynx/lynx`, `@lynx/primjs` and the four `lynx_*_service` ohpm packages to `4.0.0` as literals, outside the manifest. I could not check what ohpm publishes and have no Harmony toolchain to verify against, so I have not touched them — someone with Harmony access should confirm whether they are current, and they would benefit from the same interpolation.

- Set the documented SDK release to `4.0.2` in `version.json`.
- Use `versionJson.LYNX_VERSION` for English and Chinese quick-start download links.

### DIFF
--- a/docs/en/guide/start/quick-start.mdx
+++ b/docs/en/guide/start/quick-start.mdx
@@ -6,6 +6,8 @@ import { QRCodeSVG } from 'qrcode.react';
 import { PackageManagerTabs, Steps } from '@theme';
 import { ChoiceTabs, Go, PlatformTabs } from '@lynx';
 import * as NextSteps from '@lynx/NextSteps';
+import { Link } from '@theme';
+import versionJson from '@site/docs/public/version.json';
 
 # Try with Explorer
 
@@ -76,7 +78,7 @@ Open up the Mac App Store, search for [Xcode](https://apps.apple.com/us/app/xcod
 
 <PlatformTabs.Tab platform="macos-arm64">
 
-Download [`LynxExplorer-arm64.app.tar.gz`](https://github.com/lynx-family/lynx/releases/download/4.0.0/LynxExplorer-arm64.app.tar.gz).
+Download <Link href={'https://github.com/lynx-family/lynx/releases/download/' + versionJson.LYNX_VERSION + '/LynxExplorer-arm64.app.tar.gz'}><code>LynxExplorer-arm64.app.tar.gz</code></Link>.
 
 Then extract the downloaded archive:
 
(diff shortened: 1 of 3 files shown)


Huxpro (changes requested):
Codex-assisted review. Blocking on the existing unresolved review thread about the new `@site/docs/public/version.json` import. `docs/` is copied into downstream consumers, `@site` resolves to the consuming repository root, and this repository's AGENTS.md explicitly says not to add new `@site/*` dependencies when the target remains inside the OSS module. Please use a relative import (for these files, `../../../public/version.json`) in both locales so the shared version source remains self-contained after the docs tree is copied.

I independently verified that Lynx 4.0.2 is a published, non-prerelease release and that the four Explorer asset names used by the quick-start links exist; the exact head's configured CI/preview checks are green. I found no other concrete blocker in this change.

Please contact the requester in person for a second opinion.

PR #1412, opened 2026-09-07
docs: update lynx version to 4.1

Kingatnuaa0528:


jkzing (approved):

Huxpro (approved):

Huxpro (changes requested):
Codex-assisted review.

There is one concrete version-metadata blocker on this head: adding 4.1 as `latest` leaves the existing 4.0 entry as `latest` too. The versions metadata is intended to have a single current release; the prior 4.0 version bump demoted 3.9 from `latest` to `previous` in the same change. Please change the 4.0 entry to `previous` when promoting 4.1.

I would *not* change the new 4.1 `docs_link` away from `/` as part of this finding: `/` matches the established routing for the current/latest release. The 4.1.0 Lynx release and the referenced Explorer artifacts are published, and PrimJS 4.1.1 is also released, so I did not find a publication/version-availability blocker.

Please contact the requester in person for a second opinion if you want to discuss the release-metadata intent.

PR #1418, opened 2026-09-08
update for Lynx 4.0 Blog

toretto-wt:
Change-Id: I8306d821d8d1b6747fee677642b8b79f6a25e788

- Publish English and Chinese Lynx 4.0 release blogs covering AI Agent Skills, generative UI, ReactLynx, debugging, CSS, elements, memory queries, and upgrades.
- Prioritize `lynx-4-0` in both blog navigation arrays.

### DIFF
--- a/docs/en/blog/_meta.json
+++ b/docs/en/blog/_meta.json
@@ -1,4 +1,5 @@
 [
+  "lynx-4-0",
   "lynx-3-9",
   "lynx-rspack-2",
   "lynx-a2ui",
--- a/docs/en/blog/lynx-4-0.mdx
+++ b/docs/en/blog/lynx-4-0.mdx
@@ -1,0 +1,244 @@
+---
+title: 'Lynx 4.0: AI-Generated UI, ReactLynx Dynamic Elements, and More Open-Source Elements'
+date: 2026-09-08
+sidebar: false
+authors: ['shiwentao', 'lynx']
+badge_text: 'Lynx 4.0 is released!'
+description: 'Lynx 4.0 brings AI Agent Skills, A2UI and OpenUI generative UI, ReactLynx dynamic elements and unified debug metadata, desktop CSS updates, more open-source Elements, and global memory queries.'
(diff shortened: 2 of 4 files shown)


Huxpro (commented):
Maintainer decision needed

Codex-assisted review. The article is otherwise scoped and internally consistent on this head: CI/deploy preview are green, the new ReactLynx APIs have existing API-reference routes, and the new Clay CSS claims match current compatibility data.

One versioning choice is ambiguous enough that I don't think I should merge or block this automatically. The Upgrade Guide is written as current actionable guidance and tells readers to update Lynx and PrimJS to `4.0.0`. However, stable Lynx `4.0.2` was published on 2026-09-04. At the same time, the current `release/4.0` docs manifest still pins both `LYNX_VERSION` and `PRIMJS_VERSION` to `4.0.0`, and #1409 is separately proposing to refresh the Lynx side to `4.0.2` while keeping PrimJS `4.0.0`.

Risk: merging as-is makes a newly published 2026-09-08 blog recommend an older Lynx patch; changing it here could instead put the blog ahead of the versioned 4.0 docs before #1409 lands. Options are (1) keep `4.0.0` deliberately as historical launch guidance, (2) recommend Lynx `4.0.2` + PrimJS `4.0.0`, ideally together with/aligned to #1409, or (3) avoid hardcoding the patch in this paragraph and point readers to the versioned integration guide for the current 4.0 patch.

Recommendation: option 3 is the lowest-drift wording unless this post is intentionally meant to preserve the exact 4.0.0 launch state. Please contact the requester in person for a second opinion before resolving this maintainer choice.

Huxpro (approved):
Reviewed the latest head. The prior version-guidance gray area is resolved by pointing readers to the versioned 4.0 integration guide and using the 4.0.x line rather than hardcoding an older patch. The previously checked API-reference and compatibility claims remain covered, EN/ZH stay aligned, and CI/deploy validation is green.

PR #1304, opened 2026-08-13
docs(start): add Lynx specs repo to iOS integration guides

gongfeng98:
- add the Lynx specs repo to iOS Podfile examples
- clarify that Lynx iOS pods are resolved from the Lynx specs repo
- add follow-up notes in DevTool and dev version integration docs

Change-Id: I8032821c9d3d4df5ba9d6c9e94490e7490279bbd

- Configure the Lynx CocoaPods specs repository in iOS Podfile examples.
- Clarify Lynx pod resolution in iOS integration guides.
- Link DevTool and development-version guides to the Lynx specs repository setup instructions.
- Apply the documentation updates to English and Chinese guides.

PR #1424, opened 2026-09-10
chore(version): record 4.1 as the latest release

jkzing:
## Summary

Lynx 4.1 ships as the new stable release, so the version list has to name it. Every deployment resolves its version dropdown by fetching this file from the developing build, and `LYNX_VERSION` / `PRIMJS_VERSION` are substituted into the Gradle and CocoaPods snippets, so both have to name a version that resolves.

4.0 moves to `previous` and takes the `/4.0` base it will be built with once release/4.0 is demoted, so its links stay version-local instead of falling through to the site root, which 4.1 owns from then on.

`current_version` stays `next`: this branch is still the developing build, and the promotion happens on release/4.1.

## Published dependency check

Every coordinate the docs substitute these two values into, checked against the registries today:

| Coordinate | Registry | 4.1.0 | 4.1.1 |
| --- | --- | --- | --- |
| `org.lynxsdk.lynx:lynx` and the other 18 Lynx artifacts | Maven Central | resolves, POM and AAR both 200 | — |
| `org.lynxsdk.lynx:primjs` | Maven Central | **absent** | resolves |
| `pod 'PrimJS'` | CocoaPods | present | present |
| `pod 'Lynx'`, `LynxService`, `LynxDevtool`, `XElement` | lynx-family/Specs | resolves | — |

`PRIMJS_VERSION` is therefore 4.1.1, not 4.1.0. PrimJS is versioned on its own cadence and the manifest has carried a different PrimJS version before, on release/3.9.

Lynx no longer publishes its iOS pods to the CocoaPods trunk; they are hosted in [lynx-family/Specs](https://github.com/lynx-family/Specs), which has 4.1.0 for all four. `Lynx/4.1.0` there pins `PrimJS/quickjs` at 4.1.1, which is where the value below comes from. The docs have to declare that source, which #1429 and #1430 do.

Note that `maven-metadata.xml` for `org.lynxsdk.lynx:lynx` has `<release>` and `<latest>` pointing at a nightly, so a check that reads only those fields concludes there is no stable 4.1.0. The `<versions>` list contains it and the artifacts are served.

## Verification

`pnpm run build` passes. The published `version.json` lists 4.1 as `latest`, integration snippets render `4.1.0` with no page left rendering `4.0.0`, and the developing base stays `/next`.

## Ordering

The dropdown starts offering `/4.1/*` as soon as this lands, and those URLs only resolve once release/4.1 is promoted and served. This should land next to the switch, the way #1161 and #1196 did for the previous two releases.

## Related Links

- #1161 — the same change for 3.9
- #1196 — the same change for 4.0
- #1425 — promotes release/4.1
- #1426 — demotes release/4.0

- Set Lynx 4.1 as the latest stable version.
- Move Lynx 4.0 to `previous` with `/4.0` documentation links.
- Keep `current_version` at `next` for the developing build.
- Update `LYNX_VERSION` and `PRIMJS_VERSION` for integration snippets.

### DIFF
--- a/docs/public/version.json
+++ b/docs/public/version.json
@@ -1,7 +1,7 @@
 {
   "current_version": "next",
-  "LYNX_VERSION": "4.0.0",
-  "PRIMJS_VERSION": "4.0.0",
+  "LYNX_VERSION": "4.1.0",
+  "PRIMJS_VERSION": "4.1.1",
   "versions": [
     {
       "type": "developing",
@@ -11,9 +11,16 @@
     },
     {
       "type": "latest",
+      "version_number": "4.1",
+      "repo_branch": "release/4.1",
+      "docs_link": "/",
+      "release_number": "4.1.0"
+    },
+    {
+      "type": "previous",
       "version_number": "4.0",
       "repo_branch": "release/4.0",
-      "docs_link": "/",
+      "docs_link": "/4.0",
       "release_number": "4.0.0"
     },
     {


Huxpro (commented):
Maintainer decision needed

The change itself looks internally consistent on this exact head: it only updates `version.json`, Lynx 4.1.0 is a published stable release, the 4.1 CocoaPods spec is published, and CI / CLA / the deploy preview are green.

The remaining risk is release ordering rather than a code defect. Merging this PR makes the shared version manifest advertise 4.1 at `/4.1` immediately, while the current production deployment still serves 4.0 and `/4.1/*` only becomes valid after #1425 is landed and the Netlify production branch is switched to `release/4.1`. The PR description calls out the same coupling.

Options: (1) merge this now and accept a transient window where the 4.1 picker target may not resolve correctly, or (2) hold it until the release operator is ready, then land #1424 immediately before #1425 / the Netlify switch. I recommend (2), keeping that window as close to atomic as possible; #1426 should remain last, after the production branch switch.

This review is Codex-assisted. Please contact the requesting maintainer in person for a second opinion before the release switch.

Huxpro (dismissed):
Supersedes my earlier gray-area comment: published-dependency validation found a concrete blocker on this head.

`LYNX_VERSION` is not only release metadata; the Android integration fragment substitutes it directly into `org.lynxsdk.lynx:lynx`, `lynx-jssdk`, `lynx-trace`, and service dependencies. This PR changes that value to `4.1.0`, but the currently indexed Maven Central artifact list for `org.lynxsdk.lynx:lynx` has no stable 4.1.0 at all: its 4.1.x entries are still nightlies, with 4.0.1 the newest stable coordinate. By contrast, the iOS Specs repo does contain `Lynx/4.1.0`, so this is specifically a published-package/platform mismatch rather than a question about whether the 4.1 release exists.

Please do not make the integration snippets render `4.1.0` until the Android 4.1.0 coordinates they reference are actually published/resolvable. Either publish/verify those coordinates first, or keep `LYNX_VERSION` at a resolvable stable Android version and update it atomically when publication completes. Please make the corresponding value in #1425 match so the main and release/4.1 manifests do not diverge during promotion.

The separate release-ordering concern from my previous review still applies after this is fixed, but it is not the blocking defect.

This review is Codex-assisted. Please contact the requesting maintainer in person for a second opinion.

jkzing:
Thanks — the published-dependency angle was the right thing to check, and it did surface a real defect, but the Android reading is inverted. I re-ran the check against both registries today.

**Android is fine.** `org.lynxsdk.lynx:lynx` has a stable 4.1.0 on Maven Central: it is in the `<versions>` list of `maven-metadata.xml`, and `lynx-4.1.0.pom` and `lynx-4.1.0.aar` both return 200. I checked all 19 Lynx coordinates the docs substitute `LYNX_VERSION` into — `lynx`, `lynx-jssdk`, `lynx-trace`, `lynx-processor`, `lynx-devtool`, the four `lynx-service-*`, and the ten `xelement-*` — and every one resolves at 4.1.0.

What likely produced the opposite reading: that file's `<release>` and `<latest>` fields point at `4.1.0-nightly.202606210627.78.gdfa91d8f`, because the nightlies were published after the stable. A check that reads only those two fields concludes 4.0.1 is the newest stable and the 4.1 entries are all nightlies, which is exactly the described symptom.

**PrimJS was genuinely broken, on Android.** `org.lynxsdk.lynx:primjs` has no 4.1.0 at all; its 4.1 line starts at 4.1.1. This PR was rendering `org.lynxsdk.lynx:primjs:4.1.0`, which does not resolve. Fixed by setting `PRIMJS_VERSION` to 4.1.1, which exists on both Maven Central and CocoaPods. Precedent for the two values differing: release/3.9 carries `PRIMJS_VERSION` `3.8.0-alpha.6`.

**iOS is the actual blocker, and it is on the Lynx pods, not PrimJS.** The CocoaPods CDN index lists no 4.1.x for `Lynx`, `LynxService`, `LynxDevtool` or `XElement` — newest stable is 4.0.2 — while `PrimJS` has both 4.1.0 and 4.1.1. So `pod 'Lynx', '4.1.0'` cannot install today. The engine's 4.1.0 GitHub release, including the iOS zips those podspecs point at, was published on 2026-09-07, so this looks like the trunk push lagging rather than a missing release.

Net effect on this PR: `PRIMJS_VERSION` corrected to 4.1.1 here and in #1425, kept atomic as you asked. `LYNX_VERSION` stays 4.1.0 and the PR should not merge until the 4.1 pods are on trunk. I have no way to publish those; someone on the iOS release side needs to confirm.

On ordering, agreed with option (2) — hold until the operator is ready, then land #1424 and #1425 back to back, switch the Netlify production branch, and demote release/4.0 in #1426 last. The PR bodies now spell that sequence out.

Separately, #1427 backports the trailing-slash Edge Function to main and #1428 cherry-picks it to release/4.1, so #1425 no longer carries that file.

Huxpro (dismissed):
Supersedes my previous publication finding on the prior head. The remaining blocker on this exact head is iOS dependency resolution.

`LYNX_VERSION` is substituted into the current iOS Podfile examples, but those examples still use the CocoaPods CDN and do not configure the Lynx Specs repository. This PR itself records that `Lynx`, `LynxService`, `LynxDevtool`, and `XElement` 4.1.0 are absent from the CocoaPods CDN. The separate Lynx Specs repository does contain `Lynx/4.1.0`, and #1304 is the open docs change that teaches users to add that source.

Please either publish the 4.1 iOS pods to the source the released docs currently use, or land #1304 (or equivalent Specs-source setup) atomically before changing `LYNX_VERSION` to 4.1.0. Otherwise this PR makes the documented iOS install path resolve to a version users cannot install. Keep #1425 aligned with the same fix so the duplicated manifests do not diverge during promotion.

CI and the deploy preview are green, but they do not validate external Pod resolution.

This review is Codex-assisted. Please contact the requesting maintainer in person for a second opinion.

jkzing:
Update on the iOS blocker: it is resolved, and the reason the pods looked missing is that Lynx no longer publishes them to the CocoaPods trunk. They live in [lynx-family/Specs](https://github.com/lynx-family/Specs), whose README states the trunk is no longer maintained.

Checked that repository today: `Lynx`, `LynxService`, `LynxDevtool`, `XElement`, `LynxBase`, `BaseDevtool` and `LynxServiceAPI` all have 4.1.0. `Lynx/4.1.0` resolves to the `Lynx-4.1.0.zip` asset of the engine's 4.1.0 GitHub release, and its `Framework` subspec depends on `PrimJS/quickjs` pinned at **4.1.1** — independent confirmation of the `PRIMJS_VERSION` correction in this PR. `PrimJS` itself is not in that repository and is still served from the CDN, where 4.1.1 exists.

So both `LYNX_VERSION` 4.1.0 and `PRIMJS_VERSION` 4.1.1 name versions that resolve, and this PR is no longer blocked on publication.

One thing the source move does require: the iOS integration docs still list only `source 'https://cdn.cocoapods.org/'`, so following them cannot install `pod 'Lynx', '4.1.0'`. #1429 adds the Specs source to the Podfile snippets on main and #1430 cherry-picks it to release/4.1. #1430 should land with the promotion, otherwise the 4.1 site publishes an iOS setup that cannot resolve.

PR #1430, opened 2026-09-10
docs(ios): point the Podfile at the lynx-family Specs source

jkzing:
## Summary

Cherry-pick of #1429 onto release/4.1. This is the release that needs it: `Lynx`, `LynxService`, `LynxDevtool` and `XElement` 4.1.0 exist only in [lynx-family/Specs](https://github.com/lynx-family/Specs), while the CocoaPods CDN still stops at 4.0.2. Without the extra source, the Podfile these docs publish cannot install the version they name.

`PrimJS` is not in the Specs repository and is still served by the CDN, so both sources stay.

## Verification

Prettier and the asset-url and case-collision checks pass. The highlighted ranges were re-derived against the rendered blocks.

## Related Links

- #1429 — the same change on main
- #1425 — the promotion, which sets `LYNX_VERSION` to 4.1.0

### DIFF
--- a/docs/en/guide/start/fragments/ios/integrating-lynx-with-existing-app-ios.mdx
+++ b/docs/en/guide/start/fragments/ios/integrating-lynx-with-existing-app-ios.mdx
@@ -34,10 +34,13 @@ The core capabilities of [Lynx Engine](/guide/spec.html#engine) include basic ca
 
 Get the latest version of Lynx from Cocoapods. Then add Lynx to your Podfile:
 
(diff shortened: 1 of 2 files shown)


jkzing:
Checked both previews, which each hold one half of the final state:

- This one renders both sources, with the pod versions still at 4.0.0 because the promotion is not on this branch yet.
- The #1425 preview renders `pod 'Lynx', '4.1.0'` and `pod 'PrimJS', '4.1.1'`, with only the CDN source because this change is not on that branch yet.

Landing this one first is what matters: with #1425 alone, the published 4.1 site would name 4.1.0 while listing only the trunk, which cannot resolve it.

Huxpro (approved):

PR #1422, opened 2026-09-09
fix(compat-data): refresh CSS documentation paths

Dugyu:
## Summary

- Upgrade `@lynx-js/css-defines` from `0.0.16` to `0.0.18`.
- Regenerate CSS compatibility data and API statistics.
- Consume the corrected `lynx_path` values after the documentation restructure.

## Impact

Stale `docs/zh` paths are removed. Route validation still reports six intentionally retained paths for pending documentation pages.

## Verification

- `pnpm --filter @lynx-js/lynx-compat-data run lint`
- `pnpm --filter @lynx-js/lynx-compat-data run test`
- `pnpm run prepare`
- `pnpm run build`

Issue: lynx-family/lynx#9015

## Related

-  lynx-family/lynx#9015
- #1337 

- Upgrade `@lynx-js/css-defines` to `0.0.18`.
- Refresh the lockfile, CSS compatibility data, and API statistics.
- Consume corrected `lynx_path` values and remove stale `docs/zh` paths.
- Retain six route-validation paths for pending documentation pages.

PR #1431, opened 2026-09-10
fix(compat-data): refresh CSS documentation paths

jkzing:
## Summary

Cherry-pick of #1422 onto release/4.1, which was cut before it landed and still resolves `@lynx-js/css-defines` at 0.0.16.

The compat table turns `lynx_path` into a link with `withBase('/' + lynxPath)`. In 0.0.16, 31 of the 189 values are still repository file paths left over from the docs restructuring — `docs/zh/api/css/properties/left` and similar — so those links point at URLs the site does not publish. 0.0.18 emits site routes for all 190.

## Verification

Applied cleanly; after it, `packages/lynx-compat-data/package.json` and `pnpm-lock.yaml` are identical to main on these paths.

- `pnpm install --frozen-lockfile` succeeds, so the lockfile matches the bumped manifest, and the installed package is 0.0.18.
- `pnpm --filter @lynx-js/lynx-compat-data run test` passes, 18 tests.
- `pnpm --filter @lynx-js/lynx-compat-data run gen-css` regenerates 205 property files and leaves the working tree clean: that output is gitignored, so nothing generated needs to be checked in here.

Comparing the two published packages directly, 0.0.16 carries 31 repository-relative `lynx_path` values and 0.0.18 carries none. The effect is visible today: `lynxjs.org/docs/zh/api/css/properties/left`, the URL the current data produces, returns 404, while the route 0.0.18 points at, `lynxjs.org/api/css/properties/left`, returns 200. The same 404 reproduces on the release/4.1 branch deploy.

Six `lynx_path` values resolve to no page in this branch's docs tree, all of them properties without a page in the repository at all. Main is in exactly the same state, so this is not something the cherry-pick introduces.

## Related Links

- #1422 — the original change on main
- https://github.com/lynx-family/lynx/issues/9015

### DIFF
--- a/packages/lynx-compat-data/package.json
+++ b/packages/lynx-compat-data/package.json
@@ -40,7 +40,7 @@
     "json-schema-to-typescript": "^15.0.2"
   },
   "devDependencies": {
-    "@lynx-js/css-defines": "0.0.16",
+    "@lynx-js/css-defines": "0.0.18",
     "ts-node": "^10.9.2",
     "typescript": "^5.6.2",
     "vitest": "^2.1.2"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -269,8 +269,8 @@ importers:
         version: 15.0.4
     devDependencies:
       '@lynx-js/css-defines':
-        specifier: 0.0.16
-        version: 0.0.16
+        specifier: 0.0.18
+        version: 0.0.18
       ts-node:
         specifier: ^10.9.2
         version: 10.9.2(@types/node@18.19.130)(typescript@5.9.3)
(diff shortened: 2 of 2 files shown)


jkzing:
Confirmed against the published artifacts rather than the source, comparing the current branch deploy with this PR's preview:

| | `lynx-compat-data/css/properties/left.json` → `lynx_path` |
| --- | --- |
| `release-4-1--lynx-doc.netlify.app` | `docs/zh/api/css/properties/left` |
| `deploy-preview-1431--lynx-doc.netlify.app` | `api/css/properties/left` |

The first is the repository file path the compat table currently turns into a link, and that URL 404s. The second is the published route, which returns 200.

Huxpro (approved):

PR #1186, opened 2026-07-17
docs: bump LynxExplorer download links to 4.0.0 in quick-start (release/4.0)

KealanAU:
Closes part of #1232. This is the `release/4.0` companion to #1185. The published quick-start at the site root is built from this branch, and it still points every LynxExplorer download at release `3.8.1` even though `4.0.0` is the current release.

- Bump macOS arm64 `LynxExplorer-arm64.app.tar.gz` link to 4.0.0 (en + zh)
- Bump macOS x86_64 `LynxExplorer-x86_64.app.tar.gz` link to 4.0.0 (en + zh)
- Bump Android `LynxExplorer-noasan-release.apk` QR link and the GitHub Release tag link to 4.0.0 (en + zh)
- Bump HarmonyOS `lynx_explorer-default-unsigned.hap` link to 4.0.0 (en + zh)

All four referenced assets exist on the 4.0.0 release. The change stays limited to the Explorer download links.

Originally opened against `release/3.9`; retargeted and rebased onto `release/4.0` now that it serves the site root. Happy to open the same change against `release/3.9` and `release/3.8` if the older docs should carry it too.

PR #1185, opened 2026-07-17
docs: bump LynxExplorer download links to 4.0.0 in quick-start

KealanAU:
Closes part of #1232. The quick-start still points every LynxExplorer download at release `3.8.1`, while `4.0.0` is the current release and `docs/public/version.json` already declares `LYNX_VERSION: 4.0.0`. This updates the download and release links so new users get the current Explorer build.

- Bump macOS arm64 `LynxExplorer-arm64.app.tar.gz` link to 4.0.0 (en + zh)
- Bump macOS x86_64 `LynxExplorer-x86_64.app.tar.gz` link to 4.0.0 (en + zh)
- Bump Android `LynxExplorer-noasan-release.apk` QR link and the GitHub Release tag link to 4.0.0 (en + zh)
- Bump HarmonyOS `lynx_explorer-default-unsigned.hap` link to 4.0.0 (en + zh)

All four referenced assets exist on the 4.0.0 release. The change stays limited to the Explorer download links; SDK dependency versions, such as those in the HarmonyOS integration fragment, are left as they are.

This PR targets `main`, which serves the `/next` docs. #1186 is the companion against `release/4.0` for the published site root.

- Update English and Chinese quick-start guides to reference LynxExplorer 4.0.0 downloads for macOS, Android, and HarmonyOS.

PR #1432, opened 2026-09-10
docs: point Quick Start at the 4.1.0 Explorer and correct the Edge Function comment

jkzing:
## Summary

Two loose ends from the 4.1 release preparation, kept as separate commits.

**Quick Start Explorer links.** The page still hands out the 4.0.0 build while describing 4.1. All four artifacts it links are published on the [4.1.0 release](https://github.com/lynx-family/lynx/releases/tag/4.1.0): both simulator archives, the Android APK behind the QR code, and the Harmony HAP. Same five substitutions per language as #1185 did for 4.0.

**Edge Function comment.** #1427 landed the trailing-slash Edge Function here with a comment telling every release branch to register its own base. That is only true for an archived release, which builds at `/x.y` and serves a version home there. The current release builds at the root, where `/x.y/*` is a compatibility redirect that a bare `/x.y` already matches — registering it there only inserts a hop before the redirect, which is what #1425 measured and dropped. Comment only, no behavior change.

## Verification

Each of the four 4.1.0 artifacts returns content over HTTP. Prettier, the asset-url check and the lynx-ui route contract check pass, and `netlify.toml` still parses with the same single `/next` entry and 29 redirects.

## Related Links

- #1185, #1186 — the same Explorer bump for 4.0
- #1427 — where the comment came from
- #1425 — where the promoted-branch behavior was measured

- Update Quick Start Lynx Explorer links and QR-code references to release 4.1.0.
- Clarify `netlify.toml` release-branch documentation without changing Edge Function behavior.

### DIFF
--- a/docs/en/guide/start/quick-start.mdx
+++ b/docs/en/guide/start/quick-start.mdx
@@ -76,7 +76,7 @@ Open up the Mac App Store, search for [Xcode](https://apps.apple.com/us/app/xcod
 
 <PlatformTabs.Tab platform="macos-arm64">
 
-Download [`LynxExplorer-arm64.app.tar.gz`](https://github.com/lynx-family/lynx/releases/download/4.0.0/LynxExplorer-arm64.app.tar.gz).
+Download [`LynxExplorer-arm64.app.tar.gz`](https://github.com/lynx-family/lynx/releases/download/4.1.0/LynxExplorer-arm64.app.tar.gz).
 
 Then extract the downloaded archive:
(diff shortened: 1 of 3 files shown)


jkzing:
Closing in favour of #1433, which carries the Explorer link bump on its own against main.

The other half of this PR, the wording fix for the Edge Function comment in `netlify.toml`, is dropped. It was comment-only with no behavior change; the code it describes is unaffected.

PR #1435, opened 2026-09-10
fix(examples): remove host copy command dependency

Dugyu:
What:
- copy generated example assets with Node filesystem APIs
- cover nested exclusions, symlinks, and shell-sensitive paths

Why:
- deploy images do not consistently provide rsync
- the Lynxtron prepare step made that host dependency part of every build

How:
- preserve recursive copying, symlink dereferencing, timestamps, and ignores
- run the generator with an empty PATH in the regression test

Test: node --test scripts/lynx-example.test.js; pnpm run prepare; pnpm run build; pnpm run format:check
Change-Id: I78ee2b040e650b69c4302e88b2f899e6d0d5fd36

### DIFF
--- a/scripts/lynx-example.js
+++ b/scripts/lynx-example.js
@@ -36,7 +36,6 @@
 
 const fs = require('fs');
 const path = require('path');
-const { execSync } = require('child_process');
 
 const currentDir = process.cwd();
 const examplesDir = path.join(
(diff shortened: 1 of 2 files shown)


Dugyu:
AGENTS: please do not auto-merge this PR, only approve it. Wait for human verification.

Huxpro (approved):
Reviewed the exact latest head. The Node `fs.cpSync` replacement preserves the previous recursive-copy semantics, symlink dereferencing, timestamps, and exclusion behavior while removing the host `rsync`/`cp` dependency. The added regression test covers nested exclusions, symlinks, shell-sensitive paths, and an empty `PATH`. CI, CLA, and the Netlify preview are green. This is scoped and low-risk.

Dugyu:
Keeping `recursive: true` is intentional. When copying a symlink to a file with `dereference: true`, `fs.cpSync` throws `ERR_FS_EISDIR` without `recursive` on both Node 22.16 and Node 24. The `linked.tsx` fixture in the regression test covers this behavior.

PR #1134, opened 2026-06-24
docs: update integrating devtool.

mitchilling:
replace preset values with bootstrap values description

Change-Id: I3613beecba0b08450fae09095e32ea8432608a91

- Update Lynx DevTool integration docs to use `DevToolSettings...bootstrap().applyDevelopmentDefaultsIfUnset()` instead of preset-value setter calls
- Clarify that `Lynx DevTool` and `Lynx LogBox` switches only take effect after DevTool reaches the `ENABLED` state across iOS, Android, and Harmony
- Align iOS and Swift initialization snippets with the new enablement sequence: advance to `ENABLED`, then enable DevTool, then enable LogBox
- Align Android “Registering DevTool Service” and “Enabling DevTool” examples to apply bootstrap defaults after `enableAllSessions()` and enable LogBox after `ENABLED`
- Update Harmony ETS initialization and requirement text to reflect the `ENABLED`-state prerequisite and the revised DevTool/LogBox enablement order

PR #1437, opened 2026-09-10
docs(start): restore the DevTool bootstrap defaults on main (partial revert of #1388)

jkzing:
## Summary

This is a **partial revert** of 50f1b99b ("fix(docs): make the DevTool integration snippets compile", #1388).

#1388 bundled two unrelated changes to `guide/start/integrate-lynx-devtool`:

1. It removed the "set devtool bootstrap defaults" step from all four snippets (ObjC / Swift / Java / Kotlin), together with the ObjC `#import <Lynx/DevToolSettings.h>`, the matching bridging-header line, the two sentences that introduced the step, and it shortened the code-fence highlight ranges. The stated reason was that the call does not build against the published **4.0.0** SDKs.
2. It dropped the `bizID:` argument from the Swift `LynxServices` call.

Only (1) was version-specific, and it no longer applies now that main documents 4.1. Change (2) is version-independent and correct, so it is kept. Reverting the commit as a whole would re-break the Swift snippet.

## Why (1) is restored

`bootstrap` / `applyDevelopmentDefaultsIfUnset` is a **4.1 API**. It was introduced into the docs by ccd0376e (#1134); the site was still rendering 4.0.0 when #1388 landed, and `docs/public/version.json` on main now has `LYNX_VERSION: 4.1.0`.

In `lynx-family/lynx` on `release/4.1` (`7c7af2a75a378a1a873106a5ab0e5a62b9c8818e`):

| File | Line | Declaration |
| --- | --- | --- |
| `platform/android/lynx_android/src/main/java/com/lynx/devtoolwrapper/DevToolSettings.java` | 106 | `public BootstrapSettings bootstrap()` |
| `platform/android/lynx_android/src/main/java/com/lynx/devtoolwrapper/DevToolSettings.java` | 182 | `public void applyDevelopmentDefaultsIfUnset()` |
| `platform/darwin/common/lynx/public/devtool_wrapper/DevToolSettings.h` | 40 | `- (void)applyDevelopmentDefaultsIfUnset` |
| `platform/darwin/common/lynx/public/devtool_wrapper/DevToolSettings.h` | 59 | `@property ... DevToolBootstrapSettings *bootstrap` |

On `release/4.0` (`c8d810d75a33995bdec523394124ebc932de5b8b`), neither `bootstrap` nor `applyDevelopmentDefaultsIfUnset` occurs in either of those two files (0 matches). That is exactly the "not in the shipped artifacts" observation #1388 made — it was true for 4.0, and it is no longer true for 4.1.

So this PR restores:

- `DevToolSettings.inst().bootstrap().applyDevelopmentDefaultsIfUnset()` in the Java and Kotlin snippets
- `[[DevToolSettings sharedInstance].bootstrap applyDevelopmentDefaultsIfUnset]` in the ObjC snippet
- `DevToolSettings.sharedInstance().bootstrap.applyDevelopmentDefaultsIfUnset()` in the Swift snippet
- the ObjC `#import <Lynx/DevToolSettings.h>` and the bridging-header line
- the paragraph introducing the step (reworded, see below)
- highlight ranges covering the restored lines

## Why the Swift `bizID` fix is kept

`platform/darwin/ios/lynx_service_api/ServiceAPI.h` is **identical on `release/4.0` and `release/4.1`**; both declare only:

```objective-c
+ (id)getInstanceWithProtocol:(Protocol *)protocol;
```

There is no `bizID` parameter in either version, so `getInstanceWith(_:bizID:)` was wrong before 4.1 and is still wrong on 4.1. The snippet therefore keeps the corrected form from #1388:

```swift
(LynxServices.getInstanceWith(LynxServiceDevToolProtocol.self) as? LynxServiceDevToolProtocol)?.enableAllSessions()
```

## Why the prose is reworded instead of restored verbatim

The removed text was two sentences; the second one read:

> In the next chapter, we will introduce these bootstrap values in detail.

(with a corresponding sentence in the zh page). That promise was never fulfilled: the only other mention of a bootstrap value on the page today is the `LoadQJSBridge` / `LoadV8Bridge` note, and there is still no section covering the bootstrap values. Restoring it verbatim would re-add a dangling pointer to a chapter that does not exist, so the paragraph now only describes what the values are for:

> DevTool Service provides some bootstrap values to control default behaviors. You can configure them as needed; the snippets below apply the development defaults to the values that have not been set.

The zh page carries an equivalent sentence in natural Chinese.

## Highlight ranges

The ObjC (`{10-19}`) and Swift (`{5-14}`) ranges are restored to their pre-#1388 values, which line up with the restored blocks. The Java/Kotlin blocks are set to `{3-10}` rather than the pre-#1388 `{3-13}`: with the restored lines those blocks are 12 and 11 lines long, so `{3-13}` ran past the end of the fence. `{3-10}` covers the same configuration lines, from "register DevTool service" through the bootstrap-defaults call.

## Validation

- `pnpm run format:check` — passes
- `cspell lint` on both changed files — 0 issues
- The API-availability conclusions come from **reading the SDK headers/sources on the two `lynx-family/lynx` release branches** cited above. **No iOS or Android app was actually compiled** as part of this change; the claim here is only that the symbols the snippets reference are declared in the 4.1 SDK and were not declared in 4.0.

## Affected pages

- `docs/en/guide/start/integrate-lynx-devtool.mdx`
- `docs/zh/guide/start/integrate-lynx-devtool.mdx`

- Restore `DevToolSettings` bootstrap defaults in Objective-C, Swift, Java, and Kotlin integration examples.
- Retain the Swift `LynxServices` correction that removes the unsupported `bizID` argument.
- Align English and Chinese DevTool documentation with Lynx 4.1.0.

### DIFF
--- a/docs/en/guide/start/integrate-lynx-devtool.mdx
+++ b/docs/en/guide/start/integrate-lynx-devtool.mdx
@@ -75,7 +75,7 @@ You can configure these switches during [Lynx Environment Initialization](/guide
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
   // ...
-  // set devtool bootstrap defaults
+  // set DevTool bootstrap defaults
   [[DevToolSettings sharedInstance].bootstrap applyDevelopmentDefaultsIfUnset];
   LynxEnv *lynxEnv = [LynxEnv sharedInstance];
   // Enable all sessions debugging for DevTool (required for Lynx DevTool Desktop connection)
--- a/docs/en/guide/start/integrate-lynx-devtool.mdx
+++ b/docs/en/guide/start/integrate-lynx-devtool.mdx
@@ -65,7 +65,8 @@ You can configure these switches during [Lynx Environment Initialization](/guide
 <Tabs groupId='integrating-lynx-with-existing-app-ios'>
 <Tab label="Objective-C">
 
-```objective-c title=AppDelegate.m {9-16}
+```objective-c title=AppDelegate.m {10-19}
+#import <Lynx/DevToolSettings.h>
 #import <Lynx/LynxEnv.h>
 #import <Lynx/LynxService.h>
 #import <Lynx/LynxServiceDevToolProtocol.h>
@@ -74,6 +75,8 @@ You can configure these switches during [Lynx Environment Initialization](/guide
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
   // ...
+  // set devtool bootstrap defaults
+  [[DevToolSettings sharedInstance].bootstrap applyDevelopmentDefaultsIfUnset];
   LynxEnv *lynxEnv = [LynxEnv sharedInstance];
   // Enable all sessions debugging for DevTool (required for Lynx DevTool Desktop connection)
   [LynxService(LynxServiceDevToolProtocol) enableAllSessions];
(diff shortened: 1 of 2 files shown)


Huxpro (approved):
Reviewed the latest head against the published Lynx 4.1.0 sources. The bootstrap APIs exist on both Android and Darwin, the snippets apply the defaults before LynxEnv initialization, EN/ZH remain aligned, and CI/deploy validation is green.

Huxpro (approved):
Re-reviewed the latest head after the comment-only capitalization fix and merge from `main`. The PR diff remains scoped to the EN/ZH DevTool integration guides; the previously verified Lynx 4.1 bootstrap API availability and initialization ordering are unchanged.

PR #1442, opened 2026-09-11
Revert "[Doc] Generate public Element compatibility data (#1381)"

jkzing:
## Summary

Reverts 5afd80db75adc0f2f8a3456790fa6a66ebf0d69f — "[Doc] Generate public Element compatibility data (#1381)" — **on `release/4.1` only**.

That commit generated public Element compatibility data, but the data set describes the **4.2** Element surface. This was confirmed with the commit author: the content belongs on `release/4.2` and should not ship from `release/4.1`.

## Why this matters on `release/4.1` specifically

`release/4.1` is served at `docs_link: "/"`, so it is the version readers land on by default. The generated data drives the per-platform availability badges rendered next to Element attributes, events and methods, which means the default site was showing availability claims for a surface that release 4.1 does not expose.

Two concrete signals that the data is ahead of this branch:

1. **Forward-looking version markers that cannot be correct on a 4.1 branch.** The data carries five entries newer than the branch itself:

   | Feature | Platform | `version_added` |
   | --- | --- | --- |
   | `common` / `enable-exposure-ui-clip` | `clay_harmony` | `4.2` |
   | `common` / `overlap` | `harmony` | `4.2` |
   | `nested-image` / `loop-count` | `android` | `4.2` |
   | `scroll-view` / `experimental-android-enable-new-overflow` | `android` | `4.2` |
   | `overlay` / `android-overlay-scope` | `android` | `4.3` |

2. **Most of the newly added element files have no documentation page on this branch.** Of the 12 element files the commit added, 11 have no corresponding rendered page on `release/4.1` (`component`, `filter-image`, `list-row`, `markdown`, `raw-text`, `refresh-header`, `scroll-coordinator-header`, `scroll-coordinator-slot`, `scroll-coordinator-slot-drag`, `scroll-coordinator-toolbar`, `viewpager-item`); only `page` does.

## What this changes

Scope is exactly the 35 paths the original commit touched, restored to their pre-5afd80db contents:

- **12 element files removed** (added by that commit): the 11 listed above plus `page.json`
- **22 element files restored** to their previous contents, including `common.json`, `view.json`, `text.json`, `image.json`, `input.json`, `list.json`, `scroll-view.json`, `video.json`, `webview.json`
- **`.gitignore`**: drops the `packages/lynx-compat-data/.generated/` entry

`35 files changed, 447 insertions(+), 21715 deletions(-)`

### Compatibility-badge impact

Tracked `__compat` feature entries across these files go from **845 back to 313** (−532). Practically, the affected Element attributes/events/methods return to the badge state they had before #1381 on this branch: those that previously had no compat entry stop rendering a platform badge row, rather than rendering an incorrect one. Element pages that do exist still build and render (verified below).

The revert applies cleanly with no conflicts, and the resulting tree for these 35 paths is byte-identical to `5afd80db^`.

## Relationship to `main` — intentional divergence

**This content is intentionally kept on `main`. This PR does not touch `main` and does not propose removing it there.** The data is correct for the 4.2 surface; it is only misplaced on the 4.1 release branch.

As a direct consequence, **`release/4.1` will diverge from `main` on these files. That divergence is the expected and intended outcome of this PR** and should not be "fixed" by re-syncing the files from `main` into `release/4.1`. Any future forward-merge or cherry-pick from `release/4.1` toward `main` must not carry this revert with it.

## Validation

All run on this branch, from a clean checkout:

| Check | Result |
| --- | --- |
| `pnpm install` (triggers root `prepare`) | pass |
| `pnpm run prepare` | pass |
| `pnpm --filter @lynx-js/lynx-compat-data run lint` (schema/data `validate.ts`) | pass |
| `pnpm --filter @lynx-js/lynx-compat-data test` (vitest) | pass — 2 files, 18 tests |
| `pnpm run build` | pass — 3198 pages, sitemap generated |

Notes:

- `prepare` regenerates compat data (`gen-css`, `lint:keys --fix`, `gen-stats`). It produced **no tracked-file modifications** after the revert, so the reverted data is self-consistent with the generators on this branch.
- Element pages backed by the reverted data still render: `api/elements/built-in/page.html` and `api/elements/built-in/overlay.html` both build.
- The build emits one pre-existing vendor warning from `vscode-languageserver-types` (`Critical dependency: require function is used in a way in which dependencies cannot be statically extracted`), unrelated to this change and present before it.
- Pre-existing and unchanged by this PR: `gen-stats` warns that 6 CSS `lynx_path` values do not resolve to a page (all `-x-*` / `text-decoration-thickness` CSS properties, not Element data).


Huxpro (commented):
Maintainer decision needed

Codex-assisted review of exact head `6293db340aea4cbd054c96c81f88ef060b29a134`.

The branch mismatch is real: #1381 introduced Element compatibility data with 4.2/4.3 entries and generated files that do not belong on `release/4.1`, and this head passes CI, Compat Data Stats, CLA, and the deploy preview.

The risk is that this is a whole-commit revert rather than a release-filtered correction. #1381 also added compatibility records whose declared support is at or before 4.1—for example `blur-view` gained `ios-user-interface-style` at iOS 4.1 and several common methods at 4.0. This revert removes those together with the future-only entries, reducing the tracked Element compatibility surface by 532 entries. The checks establish that the reverted dataset is structurally valid; they do not establish that every removed <=4.1 support statement should disappear from the 4.1 docs.

Options: (1) intentionally freeze `release/4.1` Element compatibility at the pre-#1381 coverage level; (2) filter the data so only future/no-route additions are removed while <=4.1 metadata is retained; or (3) regenerate from an explicitly pinned 4.1 source if one is available.

Recommendation: prefer a 4.1-pinned or filtered regeneration over the wholesale revert unless the intended policy is explicitly to restore the pre-#1381 coverage level. I do not have a concrete defect strong enough to request changes.

Please contact me in person for a second opinion before deciding.

PR #1457, opened 2026-09-11
test(examples): enforce portable generator tooling (#1439)

Dugyu:
What:
- run the Lynx example generator regression test in CI before dependency installation
- document portable build tooling and explain the recursive copy option

Why:
- prevent undeclared host commands such as rsync from breaking Cloudflare Pages and downstream builds

How:
- expose test:lynx-example and rely on its empty PATH coverage
- require Node APIs, declared dependencies, or explicitly provisioned tools

Test: pnpm run test:lynx-example; pnpm exec prettier --check AGENTS.md CONTRIBUTING.md package.json .github/workflows/ci.yml scripts/lynx-example.js; pnpm run prepare; pnpm run build Change-Id: I1ae9a3379705942e3514012a184a786091fe185c (cherry picked from commit 1418b3ef1a2a944815c7f262466e252ca921d41f)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,9 @@ jobs:
       - name: lynx-ui Route Contract Check
         run: node scripts/ci/check-lynx-ui-routes.mjs
 
+      - name: Lynx Example Generator Test
+        run: pnpm run test:lynx-example
+
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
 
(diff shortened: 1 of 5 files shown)


Huxpro (approved):
Reviewed the exact latest head. This is a clean release/4.1 backport of the already-landed portable generator CI guard: the workflow runs the dependency-free Node regression test before install, the package script targets the existing test file, and the portability guidance/comments match the implementation. CI and deployment checks are green, and there are no unresolved review threads.

Issue #1438, opened 2026-09-10
Hardcoded SDK versions in the HarmonyOS guide and quick-start do not track LYNX_VERSION


jkzing:
## Summary

Two places in the guides pin Lynx versions as **string literals** instead of the `{versionJson.*}` placeholders. They do not follow `LYNX_VERSION`, so they silently go stale on every release. `main` is now on `LYNX_VERSION: 4.1.0` / `PRIMJS_VERSION: 4.1.1` (a15bbc4c), but both still say `4.0.0`.

One of them is not merely stale: `@lynx/xelement_svg: "3.6.0"` is **no longer published on ohpm** and cannot be installed at all.

This is a different failure mode from #1436 / #1437. Those restore docs that were correct but were "fixed" against a stale rendered version. This one is about version numbers that never tracked `LYNX_VERSION` in the first place.

## 1. HarmonyOS integration guide

`docs/{en,zh}/guide/start/fragments/harmony/integrating-lynx-with-existing-app-harmony.mdx`
(en lines 26-27, 38-43, 111; zh lines 24-25, 36-41, 109)

| Package | In docs | Published on ohpm | Note |
|---|---|---|---|
| `@lynx/lynx` | 4.0.0 | 4.1.0 | |
| `@lynx/lynx_devtool` | 4.0.0 | 4.1.0 | |
| `@lynx/lynx_devtool_service` | 4.0.0 | 4.1.0 | |
| `@lynx/lynx_http_service` | 4.0.0 | 4.1.0 | |
| `@lynx/lynx_log_service` | 4.0.0 | 4.1.0 | |
| `@lynx/primjs` | 4.0.0 | **4.1.1** | no 4.1.0 exists; 4.1 line starts at 4.1.1 |
| `@lynx/xelement_svg` | **3.6.0** | 4.1.0 | **3.6.0 is unpublished — oldest available is 3.8.1** |

The `@lynx/primjs` situation mirrors Maven Central and matches the reasoning already recorded in a15bbc4c for `PRIMJS_VERSION`.

This file uses **zero** `versionJson` placeholders, while 8 other files under `docs/en` use `{versionJson.LYNX_VERSION}`. The Android guide already does the right thing, including `implementation "org.lynxsdk.lynx:primjs:{versionJson.PRIMJS_VERSION}"`.

[...]

PR #1433, opened 2026-09-10
docs: bump LynxExplorer download links to 4.1.0 in quick-start

jkzing:
## Summary

Quick Start hands out a pre-built Explorer, and it still points at the 4.0.0 build. All four artifacts the page links are published on the [4.1.0 release](https://github.com/lynx-family/lynx/releases/tag/4.1.0): both simulator archives, the Android APK behind the QR code, and the Harmony HAP. Five substitutions per language, the same set #1185 made for 4.0.

Draft on purpose: this is release-timing material, to be landed when the 4.1 rollout is ready.

## Verification

Each of the four 4.1.0 artifacts returns content over HTTP. Prettier and the asset-url and case-collision checks pass.

## Related Links

- #1185, #1186 — the same bump for 4.0
- #1424 — records 4.1 on main

PR #1463, opened 2026-09-11
docs(harmony): track LYNX_VERSION in the ohpm dependencies

jkzing:
## Summary

Closes the first half of #1438. The HarmonyOS integration guide pinned every Lynx package as a string literal, so its snippets fell behind the moment a release shipped: they still said 4.0.0 while the site builds 4.1.0. Switching to `{versionJson.LYNX_VERSION}` and `{versionJson.PRIMJS_VERSION}` keeps them in step, the way the Android guide already works.

Two details the placeholders also settle:

- **PrimJS takes its own value.** It is versioned on its own cadence and ohpm has no `@lynx/primjs` 4.1.0 — its 4.1 line starts at 4.1.1, mirroring Maven Central and CocoaPods. `PRIMJS_VERSION` is already 4.1.1, so the snippet now renders an installable version instead of one that never existed.
- **`@lynx/xelement_svg` was broken, not just stale.** It named 3.6.0, which is not on ohpm at all; the oldest published version is 3.8.1. Anyone following that step got an unresolvable dependency. It now renders 4.1.0, which is published.

## Verification

Queried the ohpm registry for every package the guide names:

| Package | 4.1.0 | Note |
| --- | --- | --- |
| `@lynx/lynx`, `lynx_devtool`, `lynx_devtool_service`, `lynx_http_service`, `lynx_log_service` | published | |
| `@lynx/primjs` | **absent** | 4.1 line starts at 4.1.1, which is published |
| `@lynx/xelement_svg` | published | 3.6.0 absent; oldest published is 3.8.1 |

`pnpm run build` passes, and the generated `guide/start/integrate-with-existing-apps.html` renders `"@lynx/lynx": "4.1.0"` through `"@lynx/lynx_log_service": "4.1.0"`, `"@lynx/primjs": "4.1.1"` and `"@lynx/xelement_svg": "4.1.0"`. Prettier and the asset-url and case-collision checks pass. The line counts inside the fences are unchanged, and each `{3-4}` / `{3-8}` highlight range was re-checked against the rendered lines.

## Not in this PR

- The Quick Start Explorer links, the other half of #1438, are already handled by #1433.
- `integrate-lynx-dev-version.mdx` has the same defect in its HarmonyOS tab (`3.5.1` literals, while the iOS and Android tabs on that page use placeholders), but it cannot be converted yet: ohpm has no `@lynx/lynx` 4.1.0-dev, so a placeholder would render a version nobody can install. Its newest `-dev` builds are 4.0.3-dev. Left alone and reported on the issue.

## Related Links

- #1438
- #1433

### DIFF
--- a/docs/en/guide/start/fragments/harmony/integrating-lynx-with-existing-app-harmony.mdx
+++ b/docs/en/guide/start/fragments/harmony/integrating-lynx-with-existing-app-harmony.mdx
@@ -23,8 +23,8 @@ The core capabilities of [Lynx Engine](/guide/spec.html#engine) include basic ca
 ```json5 title=oh-package.json5 {3-4}
 "dependencies": {
   "@ohos/imageknife": "3.2.6",
-  "@lynx/lynx": "4.0.0",
-  "@lynx/primjs": "4.0.0",
+  "@lynx/lynx": "{versionJson.LYNX_VERSION}",
+  "@lynx/primjs": "{versionJson.PRIMJS_VERSION}",
 },
 ```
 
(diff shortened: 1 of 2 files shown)


Huxpro (commented):
Maintainer decision needed

Evidence: the EN/ZH changes are aligned; `main` resolves `LYNX_VERSION` to 4.1.0 and `PRIMJS_VERSION` to 4.1.1; the exact-head CI/build, CLA, CodeRabbit, and Netlify checks pass. The Lynx `release/4.1` source also defines these Harmony packages using the release version parameter, and its Explorer uses PrimJS 4.1.1. However, these are actionable OHPM install coordinates, and I could not independently verify every substituted package/version pair against the live OHPM registry from this review environment.

Risk: if even one 4.1.0 coordinate is not actually published, the integration snippet will fail for users despite the docs build passing.

Options: (1) verify all listed package/version pairs in the live OHPM registry and merge as-is, or (2) keep explicit versions for any coordinate that cannot be verified independently.

Recommendation: hold the merge until the live registry existence of every substituted coordinate is confirmed. This is not enough evidence to request code changes because the patch and release-source metadata are internally consistent.

PR #1464, opened 2026-09-11
docs(harmony): track LYNX_VERSION in the ohpm dependencies

jkzing:
## Summary

Cherry-pick of #1463 onto release/4.1, which is the branch serving the released site, so it is where the stale and unresolvable pins are actually visible to users.

This branch's manifest is `LYNX_VERSION: 4.1.0` / `PRIMJS_VERSION: 4.1.1`, so the snippets render the versions published on ohpm for this release, including `@lynx/primjs` 4.1.1, which has no 4.1.0, and `@lynx/xelement_svg` 4.1.0, replacing a 3.6.0 pin that is not on ohpm at all.

## Verification

The two files are byte-identical to main before the change, so the cherry-pick is clean. Registry checks and the build verification are recorded in #1463; the rendered values here are the same because both branches carry the same two version fields.

## Related Links

- #1463 — the same change on main
- #1438

### DIFF
--- a/docs/en/guide/start/fragments/harmony/integrating-lynx-with-existing-app-harmony.mdx
+++ b/docs/en/guide/start/fragments/harmony/integrating-lynx-with-existing-app-harmony.mdx
@@ -23,8 +23,8 @@ The core capabilities of [Lynx Engine](/guide/spec.html#engine) include basic ca
 ```json5 title=oh-package.json5 {3-4}
 "dependencies": {
   "@ohos/imageknife": "3.2.6",
-  "@lynx/lynx": "4.0.0",
-  "@lynx/primjs": "4.0.0",
+  "@lynx/lynx": "{versionJson.LYNX_VERSION}",
+  "@lynx/primjs": "{versionJson.PRIMJS_VERSION}",
 },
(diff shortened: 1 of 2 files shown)


Huxpro (commented):
Maintainer decision needed

Evidence: this release/4.1 backport is aligned across EN/ZH; the branch resolves `LYNX_VERSION` to 4.1.0 and `PRIMJS_VERSION` to 4.1.1; exact-head CI, CLA, and Netlify pass. The Lynx `release/4.1` source defines the corresponding Harmony packages through the release version parameter and uses PrimJS 4.1.1. However, because these snippets are actionable OHPM dependency coordinates, I could not independently verify every substituted package/version pair against the live OHPM registry from this review environment.

Risk: an unpublished 4.1.0 coordinate would make the release-branch integration instructions fail even though the documentation build succeeds.

Options: (1) verify all listed pairs in live OHPM and merge this backport, or (2) keep explicit versions for any coordinate that cannot be independently verified.

Recommendation: hold until live registry existence is confirmed. The patch itself is internally consistent, so I do not think Request Changes is justified on the evidence available.

PR #1483, opened 2026-09-14
chore(release): serve 4.2 as the production documentation

jkzing:
## Summary

Prepare release/4.2 to serve the production documentation at `/`:
- Set the current documentation version to 4.2, retain `/next` for main,
  and list 4.1 at `/4.1`.
- Proxy `/next` and historical versions to their branch deployments;
  redirect `/4.2/*` permanently to the canonical root URLs.
- Normalize the newly archived `/4.1` home before its proxy strips the
  prefix, using the existing no-loop Edge Function.
- Select the production Algolia index `lynx`.
- Include the version base in sitemap URLs, matching the main change.
  The production root URLs themselves remain unchanged.

## SDK release status

Keep the verified Lynx 4.1.0 / PrimJS 4.1.1 SDK combination. There is no
verified stable 4.2.0 release in GitHub, Maven Central, or official
podspec inventories; 4.2 nightly/alpha podspecs do not establish a stable
cross-platform SDK release. Omit the new documentation entry's
`release_number` rather than publish a nonexistent release URL.
LynxExplorer downloads are unchanged.

## Verification

- `HUSKY=0 pnpm run prepare`
- `pnpm run format:check`
- `pnpm run build`
- Case-collision, asset-URL, and lynx-ui route-contract checks.
- Cross-branch version/base/index and TOML route assertions.
- Actual Edge Function slash/query/no-loop checks.
- Generated version JSON, EN/ZH page assets and canonicals, and sitemap
  URLs checked against the intended build base.

## Related PRs

- #1482: shared version list and sitemap change; land on main first.
- #1484: archived 4.1 origin; prepare before publishing this gateway.

## Manual deployment gates

This is a draft. A human must review, mark ready, and merge it after the
main version-list change. No automated merge or platform write is used.

Before production publication, a maintainer must:
1. Verify the `release-4-2--lynx-doc.netlify.app` preview.
2. Keep a known-good production deploy available and hold automatic
   publication during the cutover.
3. Change Netlify's production branch to `release/4.2` before allowing
   the 4.1 archive changes to deploy. Wait for the archived 4.1 branch
   deployment and all historical aliases to be ready.
4. Publish a fresh 4.2 production build, clearing cache if necessary.
5. Update Algolia's root corpus to 4.2 and crawl `/4.1/` into `lynx_4_1`;
   keep development content in `lynx_next`.
6. Verify root/EN/ZH pages, assets, version switching, legacy links,
   `/next`, `/4.1`, historical versions, and search result URLs.

[...]

PR #1484, opened 2026-09-14
chore(release): archive 4.1 documentation under its version path

jkzing:
## Summary

Archive release/4.1 after 4.2 becomes the production documentation:
- Set 4.1's documentation base to `/4.1` and synchronize the version list.
- Replace root-site cross-version proxies with a local `/4.1/*` rewrite,
  and register the existing slash-normalization Edge Function at `/4.1`.
- Keep legacy tutorial, Quick Start, and lynx-ui redirects within 4.1.
  The production proxy strips the prefix before reaching this branch,
  so origin redirects must return `/4.1`-prefixed locations. Add matching
  rules for direct version-prefixed branch requests.
- Adapt and extend the existing route-contract check for those rules and
  preserve Quick Start's precedence over the general lynx-ui wildcard.
- Use `lynx_4_1` for Algolia search and include `/4.1` in sitemap URLs.

Keep the existing Lynx 4.1.0 / PrimJS 4.1.1 SDK pins and 4.1 release link.
The 4.2 documentation entry intentionally has no stable SDK release link:
no stable 4.2.0 release has been verified. Do not substitute nightly or
alpha artifacts for that missing release.

## Verification

- `HUSKY=0 pnpm run prepare`
- `pnpm run format:check`
- `pnpm run build`
- Case-collision, asset-URL, and extended route-contract checks.
- Cross-branch metadata, search-index, TOML routing, and Edge assertions.
- 84 legacy-route cases covering EN/ZH HTML/Markdown/bare tutorial paths,
  production proxy requests, and direct branch URLs without loops.
- Generated HTML asset/canonical prefixes, version JSON, and sitemap
  URLs checked against `/4.1`.

## Related PRs

- #1482: shared version list and sitemap change; land on main first.
- #1483: release/4.2 production gateway for this archived origin.

## Manual deployment gates

Do not publish these changes while release/4.1 is still Netlify's
production branch. It would remove the root site's cross-version routes.

A human must review and manually merge this draft as part of the staged
4.2 cutover, after the shared main changes have landed. Keep the known-good
production deploy held, switch Netlify's production branch to release/4.2,
and let this archive build finish as a branch deployment before publishing
the new 4.2 gateway. Verify both root and `/4.1/` branch-preview paths.

Before enabling archived search, create `lynx_4_1` from the existing index
settings, clear only the new index's copied records, and configure its
crawler for `/4.1/` and `/4.1/sitemap.xml`. Crawl and verify version-local
result URLs after those production routes are available. Never clear the
existing production `lynx` index as part of archive provisioning.

No Netlify or Algolia control-plane writes are performed by this PR.
Shared exports/aliases are unchanged; validate consuming sites against
the updated OSS pin before their scheduled sync.


Huxpro (commented):
## Maintainer decision needed

Codex-assisted review of exact head `2c395c374cf44a9adcf1ac1d96579dae339af62b`.

Evidence: the archive-side changes are coherent with the companion 4.2 cutover. `version.json` keeps 4.1 as the current archived build under `/4.1`, search moves to `lynx_4_1`, root-level legacy routes redirect into `/4.1/...`, and the final local `/4.1/* -> /:splat` rewrite supports both direct archive access and the production proxy that strips the prefix. The route-contract test is updated to exercise the proxied-origin and direct-prefixed forms. The sitemap base change is consistent with the pinned Rspress sitemap plugin. Exact-head CI and the Netlify deploy preview are green.

Risk: this branch is still the current production documentation source until the 4.2 production switch. Publishing the archive routing before that switch would intentionally change root/version behavior on the live 4.1 site. The diff is therefore technically plausible but only safe at the correct point in the release sequence.

Options: (1) merge/publish this only after `release/4.2` has become production and the 4.1 archive endpoint is ready; or (2) explicitly accept changing the current production 4.1 routing ahead of the cutover. My recommendation is option 1. This is part of an important release cutover, so I am not merging it without maintainer approval.

PR #1492, opened 2026-09-15
Update for 3.9 blog

toretto-wt:
@coderabbitai summary


Huxpro (changes requested):
Codex-assisted review.

Blocking issue: this Blog newly presents `<image>` `region-to-decode` as a Lynx 3.9 API and links readers to `/api/elements/built-in/image`, but on the target `release/4.0` branch that API reference does not document `region-to-decode` at all, and `packages/lynx-compat-data/elements/image.json` has no corresponding compatibility entry. The runtime does contain `region-to-decode`, so this is an API-reference/compatibility coverage gap exposed by this PR rather than a broken implementation claim.

Please align the release post atomically: either add the EN/ZH API-reference coverage plus validated compatibility metadata for `region-to-decode`, or remove/defer this release-note item until that reference exists. The current green build only proves the route compiles; it does not make the linked API discoverable on that page.

This review is Codex-assisted. Please contact Huxpro in person for a second opinion if you think the Blog should intentionally ship ahead of the API reference.

Huxpro (changes requested):
Codex-assisted review of the latest head.

The previous `region-to-decode` API-reference blocker is fixed, but the newly regenerated `packages/lynx-compat-data/api-stats.json` is now stale against the current `release/4.0` target. The target branch at `39501c1232fda85d314c6bd38452dabc6280a65f` contains `css/at-rule/media-query.json` and its generated stats report 2 CSS at-rules; this head does not contain that target-branch file and its regenerated stats report only 1 CSS at-rule. The overall API total staying flat while this PR adds `region-to-decode` is another symptom of the same drift.

Please update/rebase this PR onto the current `release/4.0` target and regenerate `api-stats.json` there, so the new image compatibility entry is added without dropping the Media Query stats that just landed on the target branch. Then rerun the target branch's required `Done` check on the exact head.

PR #1501, opened 2026-09-16
docs(blog): drop the 3.9 post from release/4.0

Huxpro:
Reverts the 3.9 blog post from this release branch. The blog is not versioned content, so a post committed here is never reachable from the 4.0 site.

## Why

`shared-route-config.ts` states the rule and implements it:

```ts
/**
 * The blog is not versioned content: release branches are cut before a
 * release post lands, and posts are never backported, so a release build's
 * own copy of the blog is missing everything published after the cut. Every
 * version therefore links to the developing version's blog, which is the
 * single source of truth.
 */
export const BLOG_BASE = toBasePrefix(developingDocsLink);   // → "/next"
export const BLOG_IS_CROSS_VERSION = BLOG_BASE !== SITE_BASE_PREFIX;
```

On this branch `version.json` has `current_version: "4.0"` and `docs_link: "/4.0"`, while the developing entry stays at `/next`, so `BLOG_IS_CROSS_VERSION` is `true` and every blog link the 4.0 build renders is rewritten to `/next/blog/...`:

| Location | Effect |
| --- | --- |
| `src/lib/utils.ts` `withBlogBase()` | prefixes every blog path with `/next` |
| `theme/index.tsx:393` | `blogHref = ${BLOG_BASE}${removeBase(normalizedHref)}` |
| `theme/hooks/use-blog-btn-dom.ts:89` | full page navigation when cross-version |
| `src/hooks/use-canonical-latest-blog.ts` | fetches `/next/rss/blog-rss.xml` |

A post committed to this branch is still built to `/4.0/blog/...`, but nothing in the 4.0 site links there. It is an orphan page that duplicates the canonical one under `/next`.

## What happened

[#1493](https://github.com/lynx-family/lynx-website/pull/1493) added `lynx-3-9.mdx` to this branch (status `added` — the file had never existed here, since the branch was cut before the post landed on `main`). The copy it added was a July draft, not `main`'s current post:

| | `main` | this branch before the revert |
| --- | --- | --- |
| title | Autolink, Web-Aligned CSS, and ReactLynx Portal | A2UI, Autolink, CSS Updates, and More |
| date | 2026-08-03 | 2026-07-10 |
| description | — | still advertises the dropped "Lynx for AI" |
| lines | 139 | 92 |

[#1494](https://github.com/lynx-family/lynx-website/pull/1494) then corrected the author list on that orphan copy. Both are reverted here rather than reconciled, because the post should not be on this branch at all.

## Changes

[...]

PR #1502, opened 2026-09-16
build: keep the blog out of release builds

Huxpro:
Stops release builds from emitting their own copy of the blog, so a post backported to a release branch has no effect. Follow-up to [#1501](https://github.com/lynx-family/lynx-website/pull/1501), which removes one such post by hand.

## Why

`shared-route-config.ts` already states the rule:

```ts
/**
 * The blog is not versioned content: release branches are cut before a
 * release post lands, and posts are never backported, so a release build's
 * own copy of the blog is missing everything published after the cut. Every
 * version therefore links to the developing version's blog, which is the
 * single source of truth.
 */
export const BLOG_BASE = toBasePrefix(developingDocsLink);
export const BLOG_IS_CROSS_VERSION = BLOG_BASE !== SITE_BASE_PREFIX;
```

Every consumer honours it — `src/lib/utils.ts` `withBlogBase()`, the theme's `Link` override, `use-blog-btn-dom.ts`, `use-canonical-latest-blog.ts` — but nothing stops the *build* from generating the local copy anyway. A release branch still carries `docs/*/blog/` from before its cut, so it ships:

- **Orphan pages** at `/<version>/blog/...` that nothing in that build links to.
- **Duplicate sitemap URLs** — a second address for every canonical post.
- **Stale content.** On `release/4.0` today, `lynx-3-7`, `lynx-3-9`, `lynx-4-0` and `lynx-a2ui` have all diverged from `main`.
- **A dead RSS feed.** `use-canonical-latest-blog.ts` reads `${BLOG_BASE}/rss/...` on every version, so the local feed is published and never read.

## Changes

`rspress.config.ts`, all three gated on `BLOG_IS_CROSS_VERSION`:

- `route.exclude` drops `**/blog/**`.
- `pluginRss` is not registered.
- `markdown.link.checkDeadLinks.excludes` switches to the predicate form.

`src/hooks/use-canonical-latest-blog.ts` — the cross-version fallback now points at the canonical blog index when there is no local post.

### On the dead-link check

Docs link to posts by their version-less path (`docs/{en,zh}/ai/index.mdx` → `/blog/lynx-a2ui`, `docs/{en,zh}/versions.mdx` → `/blog/lynx-open-source-roadmap-2026`) and the theme re-bases them at render time, but the build-time check validates the *authored* href against the local route set. My first attempt added glob strings to the `excludes` array and the build failed on all four files. Rspress matches an array as a `Set` of exact strings:

```js
const excludesUrl = Array.isArray(excludes) ? new Set(excludes) : void 0;
const excludesFn = 'function' == typeof excludes ? excludes : void 0;
return (url) => Boolean(excludesUrl?.has(url) || excludesFn?.(url));
```

So a subtree can only be covered by the predicate form, which the type also allows (`excludes: string[] | ((url: string) => boolean)`). Worth noting separately: the pre-existing `'/Components/Components/**'` entry can never match anything under those semantics. Left as-is here.

### On the hero badge

`useLatestBlog` fed the fallback from bundled page data. With the blog excluded there is none, and `use-blog-btn-dom.ts:84` is `if (!blogLink) return;` — the badge would sit inert whenever the `/next` feed is slow or unreachable. It now falls back to the canonical blog index. Text still comes from the SSG copy, which pairs correctly with a generic destination.

## Validation

Full production build (`pnpm build`) with `current_version` set to `4.0` to simulate a release branch:

| Check | Result |
| --- | --- |
| Build | clean, `EXIT=0` |
| Blog pages emitted | 0 |
| `doc_build/rss/` | absent |
| `/blog/...` in sitemap | none, of 3172 URLs |

`/help/blog` still builds, confirming the pattern only matches the blog directories. Prettier clean; `tsc --noEmit` reports nothing outside the pre-existing `docs/public/lynx-examples/**` fixture errors.

On the developing build `BLOG_IS_CROSS_VERSION` is `false`, so every branch added here evaluates to today's values and this is a no-op — which is also all CI on this PR can exercise.

## Open question for review

Release builds currently publish `/<version>/blog/...` and those URLs are in the deployed sitemaps, so this turns them into 404s. A 301 to the canonical post would be better for anything already indexed, but I could not tell from the repo whether cross-version paths are governed by `main`'s `netlify.toml` (it carries `/3.9/lynx-ui/*` style redirects) or by each release branch's own copy (the header comment says a release branch "registers it here alongside `/next`"). Happy to add the redirects once someone confirms which file owns them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0195V7UWUf1hYLSkMHnsqRPk

---
_Generated by [Claude Code](https://claude.ai/code/session_0195V7UWUf1hYLSkMHnsqRPk)_

- Exclude blog routes and RSS feeds from cross-version release builds.
- Apply conditional dead-link exclusions to version-less blog links.
- Fall back to the canonical blog index when no local post exists.
- Preserve developing-build behavior.
- Verify no local blog pages, RSS output, or blog sitemap entries
  with `current_version` set to `4.0`.
- Consider redirects for previously indexed versioned blog URLs.


Huxpro (commented):
Maintainer decision needed

Codex-assisted review.

Evidence: the production site currently serves versioned blog URLs such as `/4.0/blog/lynx-3-7`, and the active `release/4.1` gateway proxies the whole `/4.0/*` namespace to the `release/4.0` deploy. This PR intentionally makes cross-version release builds stop emitting their local `blog/**` pages. Therefore, once this change is propagated to an archived release branch, those already-published versioned blog URLs will become 404s unless the production gateway adds higher-priority redirects to the canonical `/next/blog/...` paths. On `main` itself `BLOG_IS_CROSS_VERSION` is false, so landing this PR does not immediately remove the current `/next` blog.

Risk: bookmarks/search-indexed historical blog URLs can break during a later release/archive rollout. Options are (1) add/coordinate gateway redirects before adopting this on each release branch, (2) deliberately accept those 404s, or (3) keep emitting the duplicate local pages. My recommendation is (1): keep the build cleanup, but make redirect coverage an explicit release gate before this reaches any cross-version release build. I would not block landing on `main` solely on this if that coordination is recorded, but I would not treat release-branch adoption as safe without it.

There is also an existing unresolved CodeRabbit finding on this exact head about `/en/blog` being accidentally exempted from dead-link checking; I am not duplicating that comment.

Please contact Huxpro in person for a second opinion.

PR #1510, opened 2026-09-17
 The Lynx 4.0 blog post needs some adjustments to its structure. Let’…

toretto-wt:
…s take it down for now and republish it once the updates are ready

Change-Id: Ic1214ee048c0919edd9aa1c738618dfb8cd9414b

## Summary

- Remove the English and Chinese Lynx 4.0 blog posts.
- Remove both posts from blog navigation metadata.
- Allow republishing after structural updates.

## Verification

- No test results supplied.


Huxpro (commented):
Maintainer decision needed

This is a Codex-assisted review. The change is internally consistent and the exact-head CI/build passes, but merging it is a substantive publication decision rather than a clearly low-risk docs cleanup.

Evidence: the Lynx 4.0 post is currently published and discoverable in both the blog index and its public route; this PR deletes the EN/ZH pages and removes both navigation entries. A repository search finds no other `lynx-4-0` route/redirect reference, so the existing published URL would no longer be generated after merge.

Risk: this temporarily retracts a release announcement that has already been public since September 11, and existing external/indexed links can become 404s until the replacement lands.

Options: (1) intentionally unpublish it and accept the temporary 404; (2) remove it from blog navigation/latest-post surfaces but keep the route alive while restructuring; or (3) add a temporary redirect/tombstone and republish later.

Recommendation: prefer (2) if the goal is only to stop promoting the post while it is being edited; choose (1) only if the route itself is intentionally being retracted. I am not blocking the PR because the deletion may be deliberate, but I would not merge this Blog change without maintainer approval.

Please contact Huxpro in person for a second opinion before landing this.

PR #1511, opened 2026-09-17
Update for 4.0 blog main

toretto-wt:
- Publish the Lynx 4.0 release post in English and Chinese.
- Feature `lynx-4-0` first in both blog listings.
- Document AI Agent Skills, A2UI, OpenUI, ReactLynx APIs,
  unified debug metadata, foldable-device support, and CSS capabilities.
- Document open-source `<webview>` and `<blur-view>` elements and
  global Lynx memory queries.
- Include examples, platform limitations, configuration requirements,
  and 4.0.x upgrade guidance.
- No test results are provided.


Huxpro (changes requested):
Codex-assisted review.

Blocking compatibility mismatch in the new `<webview>` section: both locales first present `<webview>` as supported across Android, iOS, HarmonyOS, macOS, and Windows, then say users can either load a URL or provide an HTML string directly. The current API reference marks the `html` attribute Android/iOS/Harmony-only, and the compatibility data has `clay_windows` / `clay_macos` as unsupported, while `src` is supported on the desktop Clay targets. As written, the release post therefore implies that the HTML-string mode works on macOS/Windows when the site's own API contract says it does not.

Please scope the HTML-string statement/example to the platforms that support `html` (in both EN and ZH), or, if desktop support has actually landed, update the authoritative API/compatibility data atomically with evidence from the implementation. The exact-head CI/build is green, but those checks do not catch this platform-contract mismatch.

Because this is a Blog release, I am not merging it. Please contact Huxpro in person for a second opinion.

Huxpro:
Addressed the platform compatibility feedback in 18667f53. Both English and Chinese posts now state that URL loading is available across the listed supported platforms, while direct HTML-string loading through the html attribute is limited to Android, iOS, and HarmonyOS. The HTML example headings and introductions carry the same scope. I also synced the branch with the latest main and kept the blog indexes in publication-date order. Full prepare and production build pass. Leaving the review state and conversation resolution to the reviewer.

Huxpro (approved):
Reviewed exact head 18667f5390de6ad811bad1ea8a9586a8052f0d53. The previous webview compatibility blocker is resolved in both locales; the release-post API links and compatibility claims checked here align with the current docs/compat data, and exact-head CI plus the deploy preview are green. Approving the content technically; because this is a Blog/release post, it still requires maintainer approval before merge.

PR #1525, opened 2026-09-19
docs: describe the Lynx build with Rsbuild and pluginLynx

upupming:
Describes the Lynx build in terms of Rsbuild and `pluginLynx` instead of Rspeedy, and moves the section from `/rspeedy/` to `/build/`.

- New guide: **pluginLynx and Rspeedy** — what each is, what differs, which to pick.
- Config snippets use `rsbuild.config.ts`, `defineConfig` from `@rsbuild/core`, `@rsbuild/core/types` and `npx rsbuild <command>`. Lynx-only options (`output.filename.bundle`, `performance.profile`) move into `pluginLynx({ ... })`.
- Pages Rsbuild already documents (styling, assets, resolve, plugin, build-profiling, use-rsdoctor) are dropped; their sidebar entries link to rsbuild.rs and the old paths redirect there. Tailwind CSS keeps its own page for `@lynx-js/tailwind-preset`; TypeScript keeps the Lynx type extensions.
- `cli` and `upgrade` move to lynx-stack.dev (lynx-family/lynx-stack#3996).
- `/rspeedy/*` redirects to `/build/*`.

## Summary

- Move Lynx build documentation from `/rspeedy/` to `/build/`.
- Document Rsbuild, `pluginLynx`, setup choices, and configuration examples.
- Preserve Lynx-specific Tailwind CSS and TypeScript guidance.
- Link shared Rsbuild topics to `rsbuild.rs`.
- Move CLI and upgrade guidance to `lynx-stack.dev`.
- Add redirects for legacy `/rspeedy/*` routes.
- Update navigation, route configuration, feature cards, blog links,
  and Open Graph metadata to use `build`.
- Add Lynx Bundle documentation and correct documented output trees.
- Remove default `environments: { lynx: {} }` examples from code-splitting
  documentation.
- Keep Rspeedy API references where they remain required.

## Verification

- No test results are provided.
- Review finding counts are unavailable.

PR #1485, opened 2026-09-14
docs: sync the API reference from lynx-stack

upupming:
Syncs the API reference pages (`api/config`, `api/react`, `api/genui`, `api/packages`) from lynx-stack (lynx-family/lynx-stack#3938) instead of generating them here, and redirects the old `/api/*`, `/rspeedy/{config,plugins,api}` and `/plugins` URLs.

- `scripts/sync-lynx-stack-docs.mjs` copies the directories listed in `scripts/lynx-stack-docs-dirs.txt`. Every package keeps its page, but the sidebar lists only the packages in `shown-packages.json` from lynx-stack.
- Until `@lynx-js/lynx-stack-docs` is published, the dependency is pinned to the pkg.pr.new build of lynx-stack#3938.

## What changes for a reader

The pages are generated by TypeDoc 0.28 with typedoc-plugin-markdown 4.12 instead of TypeDoc 0.26 / API Extractor, so some rendering differs from the reference on lynxjs.org today.

More than before:

- Every public package of Lynx Stack has a page (46 instead of 8), each with npm, source and changelog links.
- A configuration reference (`api/config`) grouped by where an option comes from: added by `pluginLynx`, an Rsbuild option whose default Lynx changes, `lynx.config.ts` only, or pass-through to Rsbuild.
- Every symbol links to its source on GitHub.
- ReactLynx is grouped into Hooks, Components, Functions and Types instead of one flat alphabetical list, with the directives, macros, global events and import attributes pages above them.
- `useErrorBoundary`, `Children`, `markFirstScreenSyncReady`, `CloneElement`, `CreateElement` and `ReactLynxChildren` are documented for the first time.

Rendering that the toolchain changed, and this PR accepts:

- Overloads and constructors are headed `Call Signature` and `Constructor` instead of repeating the name.
- A function-typed property no longer carries `()` in its heading.
- The source of a symbol is a line under its signature instead of a `Defined in` section at the end.
- `Inherited from` renders as a code block.
- `@alpha` and `@beta` render as a tag instead of the API Extractor warning paragraph.
- Every exported symbol has its own page, so a package is a directory instead of one long page.

## Sidebar

- A package lists what you call - classes, functions, variables, namespaces - with no kind directory in between. The types they refer to are reached from the pages that use them.
- The section is `Build`, holding `Config` and `Plugins and tools`. Every directory starts collapsed.
[...]

Issue #1320, opened 2026-08-18
fix(docs): preserve lynx-ui aliases across versioned website routes


f0rdream:
## Background

The current and `next` lynx-ui routes work at `/ui`, and unversioned `/lynx-ui` aliases redirect correctly. Historical deployments use different route names, however:

- 3.8 serves `/3.8/lynx-ui/*`; `/3.8/ui/*` currently returns 404.
- 3.9 serves `/3.9/ui/*`; `/3.9/lynx-ui/*` currently loses the version by redirecting to unversioned `/ui/*`.
- Versions 3.4 through 3.7 predate the website-hosted lynx-ui docs and have neither route.

This is separate from #1310 because it concerns historical version routing and production-layer behavior.

## Scope

- Redirect `/3.8/ui/*` to `/3.8/lynx-ui/*`, preserving locale and path.
- Redirect `/3.9/lynx-ui/*` to `/3.9/ui/*`, preserving locale and path.
- Cover English and Chinese, bare roots, `.html`, `.md`, query strings, and fragments.
- Retain current unversioned and `/next` redirect behavior.
- Add a route-contract CI check for canonical and compatibility mappings.
- Apply the change at the production routing layer or backport it to the relevant release deployments; changing only `main` is insufficient if historical artifacts own those paths.

## Acceptance criteria

- Current, `next`, 3.8, and 3.9 route matrices all resolve or redirect to the correct version-local content.
- No compatibility redirect silently changes the selected documentation version.
- CI detects removal or regression of any required mapping.

PR #1530, opened 2026-09-20
docs: describe the Lynx build with Rsbuild and pluginLynx

upupming:
Describes the Lynx build in terms of Rsbuild and `pluginLynx` instead of Rspeedy, and moves the section from `/rspeedy/` to `/build/`.

- New guide: **pluginLynx and Rspeedy** — what each is, what differs, which to pick.
- Config snippets use `rsbuild.config.ts`, `defineConfig` from `@rsbuild/core`, `@rsbuild/core/types` and `npx rsbuild <command>`. Lynx-only options (`output.filename.bundle`, `performance.profile`) move into `pluginLynx({ ... })`.
- Pages Rsbuild already documents (styling, assets, resolve, plugin, build-profiling, use-rsdoctor) are dropped; their sidebar entries link to rsbuild.rs and the old paths redirect there. Tailwind CSS keeps its own page for `@lynx-js/tailwind-preset`; TypeScript keeps the Lynx type extensions.
- `cli` and `upgrade` move to lynx-stack.dev (lynx-family/lynx-stack#3996).
- `/rspeedy/*` redirects to `/build/*`.

## Summary

- Move Lynx build documentation from `/rspeedy/` to `/build/`.
- Document Rsbuild, `pluginLynx`, setup choices, and configuration examples.
- Preserve Lynx-specific Tailwind CSS and TypeScript guidance.
- Link shared Rsbuild topics to `rsbuild.rs`.
- Move CLI and upgrade guidance to `lynx-stack.dev`.
- Add redirects for legacy `/rspeedy/*` routes.
- Update navigation, route configuration, feature cards, blog links,
  and Open Graph metadata to use `build`.
- Add Lynx Bundle documentation and correct documented output trees.
- Remove default `environments: { lynx: {} }` examples from code-splitting
  documentation.
- Keep Rspeedy API references where they remain required.

## Verification

- No test results are provided.
- Review finding counts are unavailable.

Supersedes #1525, recreated from a branch in this repository so the stack can include it.


Huxpro (dismissed):
Codex-assisted review on exact head `2aa2e4cff02c89523992d25ca662dd1f44ac69eb`.

Requesting changes for two concrete compatibility regressions that remain on the latest head:

1. **The recommended plain-Rsbuild example currently depends on unreleased behavior.** The guide recommends Rsbuild + `pluginLynx` and shows `pluginReactLynx()` without an explicit `environments: { lynx: {} }`. However, the currently released `@lynx-js/rsbuild-plugin` line is still `0.1.2`, while lynx-stack#3954 explicitly documents that, without `environments`, Rsbuild falls back to the `web` environment and Lynx-specific behavior is skipped. The fix is still represented by an unconsumed patch changeset on lynx-stack `main`. Please either restore the explicit Lynx environment / add a precise minimum-version caveat until the patch is published, or land this after the corresponding package release and verify the published version.

2. **Preserve the old public API routes.** This PR moves the generated API tree from `/api/rspeedy/*` to `/api/build/*`; the latest commit correctly updates internal references, but there are still no compatibility redirects for the old `/api/rspeedy/*` URLs. Since the old API directory is removed, existing bookmarks and external links will 404. Please add redirects for the old API namespace, including the applicable locale/version-prefixed forms, and validate the route contract.

Exact-head CI is green, including the full build, so these are public documentation/package-compatibility and route-compatibility defects outside the current checks.

Please contact Huxpro in person for a second opinion.

upupming:
@Huxpro For 1, We will release it soon alongside this Docs PR

i18n.json
```diff
@@ -239,10 +239,6 @@
     "en": "ReactLynx",
     "zh": "ReactLynx"
   },
-  "nav.rspeedy": {
```
Dugyu, line 242: Let's add `nav.build`?

Dugyu (changes requested):
Request changes to block accidental merge by automated agents.

netlify.toml
Dugyu, line 1: For redirect contract, could we also make it explicitly version-aware? Since the current stable site corresponds to `4.1`, while main is published under `/next`:

- `/next/rspeedy/*` may redirect to the corresponding `/next/build/*` routes.
- Unversioned `/rspeedy/*` should redirect to `/build/*` when this documentation becomes the stable version.
- /`4.1/rspeedy/* `and earlier versioned routes should continue serving their version-local Rspeedy documentation.
- Legacy pages should redirect to `/build/*` only when the content remains semantically equivalent; otherwise they should point to the canonical Rspeedy documentation or a dedicated migration guide.

docs/en/build/tools.md
Dugyu, line 1: The new “Rsbuild and Rspeedy” page provides a clear conceptual comparison and selection guide. Could we also provide an explicit migration path for existing Rspeedy projects, either as a section on this page or as a dedicated guide linked from it?

A migration section could cover:
- dependency and `package.json` script changes (`rspeedy` → `rsbuild`)
- `lynx.config.ts` → `rsbuild.config.ts` and the new defineConfig import
- `source.entry` string → object form
- moving Lynx-specific options into `pluginLynx({ ... })`, such as `output`
- TypeScript client types
- behavioral/default differences and minimum supported versions
- a complete before/after example and verification steps

The existing `rspeedy/upgrade` target upgrades dependencies within Rspeedy, so it does not cover this Rspeedy → Rsbuild transition.

Dugyu (commented):
The `/rspeedy` → `/build` transition is clear for new users, but existing Rspeedy users may lose their path forward. I left two comments about preserving that continuity:

1. Provide an explicit Rspeedy → Rsbuild migration guide for existing projects.
2. Keep redirects version-aware, following the existing `lynx-ui` precedent, so older versions retain their version-local Rspeedy docs.

Huxpro (dismissed):
Codex-assisted review on exact head `2aa2e4cff02c89523992d25ca662dd1f44ac69eb`.

Requesting changes for two concrete defects that remain after re-validating the current head:

1. **The recommended plain-Rsbuild example still depends on unreleased behavior.** `docs/en/build/tools.md` recommends Rsbuild + `pluginLynx` and shows `pluginReactLynx()` without an explicit `environments: { lynx: {} }`. Upstream lynx-stack#3954 documents that, without `environments`, Rsbuild otherwise falls back to the `web` environment and Lynx-specific behavior is skipped. That fix is merged upstream, but `@lynx-js/rsbuild-plugin` is still version `0.1.2` on current `lynx-stack/main`, so the documentation is ahead of the released package contract. Please either restore the explicit Lynx environment / add an exact minimum-version caveat until the patched package is released, or land this after the corresponding release and verify the published version.

2. **The scheduled API regeneration path was not migrated with the API namespace.** This PR moves the checked-in generated reference from `docs/{en,zh}/api/rspeedy/*` to `docs/{en,zh}/api/build/*`, but `.github/workflows/update-api-docs.yml` still runs `scripts/update-api-docs.sh`, and that script still calls `sync_api_extractor_docs "rspeedy" "rspeedy" "true"`. After this PR lands, the daily regeneration workflow will therefore recreate/update `docs/{en,zh}/api/rspeedy/*` instead of the new `api/build` tree, causing the generated reference to diverge immediately. Please migrate the generator target (and its workflow/docs wording) to the new namespace and validate a regeneration against the current head.

I rechecked the old-route concern from my previous review and dismissed that review: `netlify.toml` already contains the `/api/rspeedy/*` → `/api/build/:splat` redirects for the unversioned, zh, `/next`, and `/next/zh` forms, so that is not a blocker.

Exact-head CI is green, but the normal PR CI does not execute the scheduled cross-repo API regeneration workflow, so it does not catch item 2.

Please contact Huxpro in person for a second opinion.

Huxpro (changes requested):
Codex-assisted review on exact head `4bfa7a72914386f047e98d256cf59623e6df0579`.

The previous API-regeneration blocker is fixed by `797c1cce`: the scheduled generator now syncs the Rspeedy API output into `api/build`, and the workflow/CODEOWNERS wording moved with it.

One concrete compatibility blocker remains: the recommended plain-Rsbuild configuration still relies on behavior that has not shipped in the stable package line. The page recommends Rsbuild + `pluginLynx` and shows `pluginReactLynx()` without an explicit `environments: { lynx: {} }`; the new migration section sends existing users down the same path. Upstream lynx-stack#3954 explicitly fixes the pre-release behavior where omitting `environments` makes Rsbuild fall back to the `web` environment and skips Lynx-specific behavior. That fix has a patch changeset for `@lynx-js/rsbuild-plugin`, but current `lynx-stack/main` still reports `@lynx-js/rsbuild-plugin` as `0.1.2`, so the docs are still ahead of the stable package contract.

Please either keep the explicit Lynx environment / document the exact minimum released version that contains #3954, or land this only after the patched package is published and verify the published version. Until then, copying the recommended example with the stable package can produce the wrong environment.

Please contact Huxpro in person for a second opinion.

i18n.json
upupming: `nav.rspeedy` had no consumers: on `main`, `git grep nav.rspeedy` matches only `i18n.json` itself, and the same holds for every other `nav.*` key (`nav.advanced`, `nav.guides`, `nav.react`, …). The subsite label comes from `shared-route-config.ts`, which this PR already updates — `{ value: 'build', label: 'Build', description: 'Build Tool for Lynx', descriptionZh: 'Lynx 构建工具' }`.

So `nav.build` would only add a second dead key. Left it out, but happy to add it if you'd rather keep the set symmetric.

netlify.toml
upupming: All four forms are covered for both namespaces: `/rspeedy/*`, `/zh/rspeedy/*`, `/next/rspeedy/*`, `/next/zh/rspeedy/*` → `/build/:splat`, and the same four for `/api/rspeedy/*` → `/api/build/:splat`. `/next/*` works on this deploy preview today, and the unversioned form takes effect when this becomes stable — exactly the two cases you list.

Versioned prefixes aren't served from `main`'s `netlify.toml`; they come from the release deploys, which already fold into the unversioned form. `https://lynxjs.org/4.1/rspeedy/index.html` 301s to `/rspeedy/index.html` right now, so once this lands the browser chains on to `/build/index.html`. No extra rule needed here.

docs/en/build/tools.md
upupming: Added, as a **Migrating an Rspeedy project** section at the end of `docs/{en,zh}/build/tools.md` (4bfa7a7): dependency swap, the CLI mapping, `lynx.config.ts` → `rsbuild.config.ts`, and the `tsconfig.json` `types` change — plus the two that actually bite, `source.entry`/`output.filename` needing the object form, and `output.filename.bundle`/`performance.profile` being Rspeedy config keys that have to move into `pluginLynx({ ... })`.

upupming:
Item 2 was real, thanks. `scripts/update-api-docs.sh` still ran `sync_api_extractor_docs "rspeedy" "rspeedy" "true"`, so the daily workflow would have recreated `docs/{en,zh}/api/rspeedy/*` right after this landed. Fixed in 797c1cc — the sync target is now `build`, with `.github/workflows/update-api-docs.yml` and the `/docs/**/api/rspeedy/` entry in `.github/CODEOWNERS` following it. `target` is that script's only path coupling, so nothing else needed to move.

On item 1: `@lynx-js/rsbuild-plugin` will be released with lynx-stack#3954, which makes `lynx` the default environment. Leaving the examples as they are.

Dugyu:
Three compatiability layers are still needed:

### 1. **Rspress client-side redirects**

The `/rspeedy/*` → `/build/*` redirects currently exist only in `netlify.toml`. They cover direct requests and page refreshes, but not SPA navigation intercepted by the Rspress router. Equivalent `pluginClientRedirects` rules are needed for SPA navigation and non-Netlify deployments (e.g., lynx-website inhouse, which won't pick any rules from netlify.toml).

Related PR: #1154

### 2. **Version-local reverse aliases**

The version switcher preserves the current page path and only replaces the version prefix:

```text
/next/build/output
→ select 3.9
→ /3.9/build/output
```

However, 3.9 published this section under `/rspeedy`, so the actual historical route is:

```text
/3.9/rspeedy/output
```

The production routing layer should therefore provide **reverse** aliases such as:

```text
/3.9/build/* → /3.9/rspeedy/*
/4.0/build/* → /4.0/rspeedy/*
/4.1/build/* → /4.1/rspeedy/*
```

Imagine when you are on the `/next/build`, then switches to `/4.0/build` by clicking the version switcher dropdown menu, suddenly the page becomes blank, as historical version cannot recognize `/4.0/build/*`. 

Since historical deployments could not have anticipated the future `/build` name, these aliases must be handled by the **current** shared Netlify routing layer **before requests reach the historical artifacts**.

This follows the precedent established by [issue #1320](https://github.com/lynx-family/lynx-website/issues/1320): route names may be translated, but the selected documentation version must remain unchanged. New pages without historical equivalents may also need explicit fallbacks to that version’s Rspeedy landing page. 

Related PR: #1321 

### 3. **Legacy subsite recognition**

PR #1160 made the `ui` subsite recognize both `ui` and its legacy `lynx-ui` segment. The same compatibility should be considered here so the `build` subsite recognizes both `build` and `rspeedy`.

This is a subtler compatibility layer. It does not perform the redirect itself, it ensures that an old `/rspeedy/*` pathname is still classified as the Build subsite and receives the correct `data-subsite="build"` theme while the client-side redirect is **being processed**.

docs/en/build/tools.md
Dugyu: Thanks! You're right about the Netlify routes. I realize I left my previous comment on `netlify.toml`, which mixed up the two deployment layers and made my concern unclear. Sorry for the confusion.

Detailed explanation has been attached to the end of this PR thread.

i18n.json
Dugyu: Although neither key currently drives the OSS navigation, and I also found no references to `nav.rspeedy` on the downstream `main` branch, `i18n.json` is shared with downstream overlays. Would it make sense to keep `nav.rspeedy` while adding `nav.build` as a low-cost compatibility measure during the transition?

I tend to view `i18n.json` as a shared translation dictionary, or a lightweight semantic contract, rather than only an inventory of strings currently rendered by the OSS site.

Huxpro (changes requested):
Codex-assisted review on exact head `bf706e3fa755bcdd7af0bd54706b51fa3e805159`.

The two latest commits correctly address the recent version-local routing and downstream-overlay feedback: historical `/build/*` paths now reverse-map to each release's `/rspeedy/*` namespace, and `i18n.json` preserves `nav.rspeedy` while adding `nav.build`.

One concrete route-compatibility defect remains in the rename itself. The `/rspeedy/*` → `/build/*` compatibility currently exists only in `netlify.toml`. `rspress.config.ts` already documents the repository's intended dual-layer contract for the previous `/lynx-ui` → `/ui` rename: Netlify handles server-side hits, while `pluginClientRedirects` keeps legacy links working for SPA-side navigation and hosts that do not consume Netlify rules. This PR does not add the equivalent client redirects for `/rspeedy` (EN/ZH), so old Rspeedy links can still fail when navigation is handled by Rspress or on non-Netlify consumers even though direct production requests redirect correctly.

Please add the equivalent `pluginClientRedirects` rules for the legacy Rspeedy namespace (including the localized form) and focused coverage for that client-side/non-Netlify contract. Exact-head CI is green, but the configured route-contract step is still the lynx-ui check and does not validate this new rename path.

I am not repeating the separately tracked published-package compatibility finding in this review.

Please contact Huxpro in person for a second opinion.

upupming:
Thanks for the detailed write-up. 2 and 3 were real; 1 belongs elsewhere.

**1. Client redirects.** The downstream deployment carries its own `rspress.config.ts` with its own `pluginClientRedirects` list and does not inherit this repo's plugin configuration, so rules added here would never reach it — we'll make that change on the internal site separately. For lynxjs.org itself, no page in this PR links to `/rspeedy/*` or `/api/rspeedy/*` any more (0 matches across `docs/` and `theme/`), so there is no in-site navigation left for the router to intercept; every remaining entry point — bookmarks, search results, links from other sites, including the `/next/rspeedy/*` form — is a hard navigation that `netlify.toml` already answers. Not doing it in this PR.

**2. Version-local reverse aliases.** Confirmed and fixed in b48ba39. `https://lynxjs.org/3.9/build/output.html` is a 404 today, exactly the blank page you describe, and every release branch from 3.4 to 4.2 publishes this section under `/rspeedy`. Following the `/3.8/ui` → `/3.8/lynx-ui` precedent, `netlify.toml` now maps `/{4.1,4.0,3.9,3.8,3.7,3.6,3.5,3.4}` (and their `/zh` forms) `/build` and `/build/*` onto the matching `/rspeedy` route — 32 rules, no duplicate `from` in the file.

**3. Legacy subsite recognition.** Done in #1485. The hard-coded `ui`/`lynx-ui` pair in `shared-subsite-routes.ts` is now a `LEGACY_SEGMENTS` map with `build: 'rspeedy'` added, so an old `/rspeedy/*` pathname still classifies as the Build subsite and keeps its theme while a redirect is being processed. Covered by a test.

i18n.json
upupming: Fair enough — treating it as a shared dictionary rather than an inventory of rendered strings makes sense for a downstream overlay. `nav.rspeedy` is restored and `nav.build` added alongside it in bf706e3.

Huxpro (dismissed):
Codex-assisted review on exact head `6fc18412ca128c70ed043c685a18495d9690eb91`.

The latest merge-from-main leaves the configured PR CI failing at `Validate → Format Check`; because that step fails, the compatibility-data checks and production Build are skipped. This is a concrete failed acceptance criterion on the current head, so it is not safe to merge yet. Please apply the formatting fix indicated by CI and rerun the configured checks through Build.

I am not repeating older compatibility or route findings in this review.

Please contact Huxpro in person for a second opinion.

